PR #57, opened 2017-08-24
fixing keyboard dragging for large items

### LINKED ISSUE #54: Keyboard reordering does not work correctly when dragging a huge item
## Bug or feature request?
Bug

### Expected behaviour

When I reorder a list with different sized items with a keyboard it should reorder correctly.

### Actual behaviour

If the item dragging is more that double the height of the next item then:
- the dragging item moves into the correct position
- the non dragging item does not move into its new location 

### Steps to reproduce

- Start a keyboard drag with an item that is more than double the height of the one next to it
- try to drag past the next item

### Demo
https://www.webpackbin.com/bins/-KsCUEpbt2X70IoQmbhj

alexreardon:
Fixes #54 

The initial commit fixes the issue, but there is some house keeping to do before this will get merged

- [x] fix broken tests
- [x] write new tests for new logic
- [x] rename `get-diff-to-jump-to-next-index` to something clearer
- [x] fix flow and linting
- [x] update story so it stops logging react warning


alexreardon:
This is good to go. However, it is the end of the day and I want to take another look at it tomorrow before merging. 

@jaredcrowe could you take a look if you can?

alexreardon:
For those who are interested here was the cause:

When we were moving something with a keyboard we calculated where it needed to move to and moved it to that new location. This would usually cause the item in the destination location to move out of the way because the center position of the dragging item goes over the edge of the non dragging item. However, if you are dragging an item that is more than double in size of the item that is not dragging, then the center position of the dragging item would not go over the edge of the non dragging item. This would mean that the non-dragging item would not be triggered to move. 

I have changed the algorithm so when you move with a keyboard it forces the non dragging item to move regardless of whether an edge was crossed.

| Before          | After           |
| ------------ | ----------- |
| large-item-bug mov-scaled-0 3 | large-item-bug-fix mov-scaled-0 3 |

Issue #65, opened 2017-08-30
Server rendering the component.


romellogood:
## Feature Request

Are there any plans to support serverside rendering of this library? Currently when rendered on the server an error is printed out.

alexreardon:
Short term: the application should be able to render out without an errors and then be useable once react kicks in on page load. 

Long term: once we figure out a programatic drag api it *might* be feasible to server side render into a mid drag (although this seems a bit strange)

Focusing on the short term then; can you please:

- confirm that you are using https://facebook.github.io/react/docs/react-dom-server.html
- provide a small case that fails and the code that drives it
- provide the error message

romellogood:
Thanks for getting back to me so fast. I can confirm that I am indeed using react-dom-server and can provide the error message below. I will try and comment with a small case that fails and the code tomorrow.

```
ReferenceError: window is not defined
at Object.<anonymous> (/Users/rgoodman/workspace/passplan-ui/node_modules/react-beautiful-dnd/lib/state/create-store.js:23:24)
at Module._compile (module.js:569:30)
at Object.Module._extensions..js (module.js:580:10)
at Module.load (module.js:503:32)
at tryModuleLoad (module.js:466:12)
at Function.Module._load (module.js:458:3)
at Module.require (module.js:513:17)
at require (internal/module.js:11:18)
at Object.<anonymous> (/Users/rgoodman/workspace/passplan-ui/node_modules/react-beautiful-dnd/lib/view/drag-drop-context/drag-drop-context.js:37:20)
at Module._compile (module.js:569:30)
```

alexreardon:
Yeah perfect - need to check we are not referencing any window globals in the initial render and then we should be good to go

alexreardon:
I'll wait for your test case so i can be sure that everything will work for you 👍

romellogood:
To make sure we are on the same page, when you say 'test case' do you mean a minimal build (backend code, frontend code, build code, etc) or just the components using the library?

alexreardon:
The best would be a full setup

romellogood:
Hey! So i created a full set up and posted it [here](https://github.com/romellogood/ssr-react-beautiful-dnd).

I took the drag and drop example that is listed in this repo's readme and copied it into the [DragNDrop.js](https://github.com/romellogood/ssr-react-beautiful-dnd/blob/master/src/DragNDrop.js).

[...]

PR #66, opened 2017-08-31
Checking that the window is defined before accessing it.

romellogood:
Should take care of issue #65 

Checking that the window is defined before accessing it so that the module doesn't break on the server.

### DIFF
--- a/src/state/create-store.js
+++ b/src/state/create-store.js
@@ -6,7 +6,7 @@ import hookMiddleware from './hook-middleware';
 import type { Store, Hooks } from '../types';
 
 // eslint-disable-next-line no-underscore-dangle
-const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
+const composeEnhancers = typeof window !== 'undefined' ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ : compose;
 
 export default (hooks: Hooks): Store => createStore(
   reducer,
--- a/src/state/create-store.js
+++ b/src/state/create-store.js
@@ -1,12 +1,14 @@
 // @flow
+/* eslint-disable no-underscore-dangle */
 import { applyMiddleware, createStore, compose } from 'redux';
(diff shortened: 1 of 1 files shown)


romellogood:
Following the fix laid out here for [universal apps](https://github.com/zalmoxisus/redux-devtools-extension#12-advanced-store-setup).

```
For universal ("isomorphic") apps, prefix it with typeof window !== 'undefined' &&.
```

alexreardon:
Looking good. I'll have a quick scan through, but i do not think we are using any other browser globals on launch.

src/state/create-store.js
```diff
@@ -1,12 +1,14 @@
...
+/* eslint-disable no-underscore-dangle */
 import { applyMiddleware, createStore, compose } from 'redux';
 import thunk from 'redux-thunk';
 import reducer from './reducer';
 import hookMiddleware from './hook-middleware';
 import type { Store, Hooks } from '../types';
 
-// eslint-disable-next-line no-underscore-dangle
-const composeEnhancers = window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ || compose;
+const composeEnhancers = typeof window === 'object'
```
alexreardon: Can you please add a comment above stating 
1. We are checking if window is available because it might not be when server side rendering
2. That this came from https://github.com/zalmoxisus/redux-devtools-extension#12-advanced-store-setup

alexreardon:
Thanks for looking into this @romellogood !!!!

alexreardon:
I have [forked your example](https://github.com/alexreardon/ssr-react-beautiful-dnd). I found it was not doing the correct SSR (it was returning `undefined` instead of the statically rendered application). I have checked my fork and everything is looking good.

alexreardon:
Once you add the comments listed above we are good to go

romellogood:
Made the changes you requested!

alexreardon:
🎉

I will try to do a release tomorrow afternoon which will contain this fix. Well done @romellogood !

romellogood:
Thanks :)

PR #67, opened 2017-08-31
fixing force press interaction

### LINKED ISSUE #62: Force press clicking messes with Dragging on Safari
## Bug or feature request?
Bug

### Expected behaviour

When releasing her finger from the trackpad, the card should remain to the list; it should not remain sticked to the cursor

### Actual behaviour

See video https://www.youtube.com/watch?v=-gWQ9hkjUj8

The radar effect happens when the user releases the card.
Even though the user releases the card, she can still move things around.

This bug manifests itself on all the example pages.

### Steps to reproduce

See https://www.youtube.com/watch?v=-gWQ9hkjUj8

You probably need a magic trackpad instead of a mouse.
Isssue manifests itself in the recent Mac OS, at the very least.

I suspect, but I haven’t time to dig in, that it might be related to “force touch”.

### Demo

Here is a video with pointer events recorded, to help you:

 https://www.youtube.com/watch?v=-gWQ9hkjUj8

alexreardon:
This change fixes #62. I can see why safari does what it does - but custom browser behaviour like things makes me super sad. 

/cc @v0lkan

- [x] write tests

### DIFF
--- a/src/view/drag-handle/drag-handle-types.js
+++ b/src/view/drag-handle/drag-handle-types.js
@@ -45,7 +45,6 @@ export type Props = {|
 |}
 
 // Custom event format for force press inputs
-export type MouseForceChangedEvent = {
-  preventDefault: () => void,
+export type MouseForceChangedEvent = MouseEvent & {
   webkitForce?: number,
 }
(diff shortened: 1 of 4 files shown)


alexreardon:
Heads up that the safari touch events do not fire correctly when in the iframe - you need to open the inner iframe in its own page to see everything working correctly

Took me a while to figure that out

PR #69, opened 2017-08-31
adding tests for server side rendering

alexreardon:
@romellogood recently added a change #65 that enabled server side rendering to work correctly. This pull request builds on that and adds some tests to ensure that we do not accidentally break server side rendering going forward.


test/unit/integration/__snapshots__/server-side-rendering.spec.js.snap
```diff
@@ -0,0 +1,5 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
alexreardon: first snapshot test!

test/unit/integration/server-side-rendering.spec.js
```diff
@@ -0,0 +1,59 @@
+/**
+* @jest-environment node
```
alexreardon: finding out that jest did this was amazing

PR #72, opened 2017-09-01
jaredcrowe/animated placeholder

jaredcrowe:


alexreardon:
Do you want to wait until we have your changes before merging this?

src/view/placeholder/AnimatedPlaceholder.jsx
```diff
@@ -0,0 +1,47 @@
+import styled, { css, keyframes } from 'styled-components';
```
alexreardon: we do not want to add a dependency on styled-components

src/view/droppable/droppable.jsx
```diff
@@ -62,14 +64,14 @@ export default class Droppable extends Component {
   }
 
   getPlaceholder() {
-    if (!this.props.placeholder) {
-      return null;
-    }
+    const isVisible = Boolean(this.props.placeholder);
```
alexreardon: as discussed - lets not animate the draggable placeholder for now

src/view/placeholder/AnimatedPlaceholder.jsx
alexreardon: Please remove any code that uses it in `/src` :)

src/view/placeholder/AnimatedPlaceholder.jsx
```diff
@@ -0,0 +1,47 @@
...
+    height: 0;
+    width: 0;
+  }
+  to {
+    height: ${height}px;
+    width: ${width}px;
+  }
+`;
+
+const animateOut = ({ height, width }) => keyframes`
```
alexreardon: We will need to use a different approach for this. We might need to create a style tag with the (prefixed) keyframe animation inside and append it / remove it from the head as needed.

alexreardon: I think we can get away with just adding a `transition: height` on the element rather than keyframes

alexreardon:
Closing  this until we discuss this further

PR #74, opened 2017-09-06
animated droppable placeholder spike

jaredcrowe:
A spike of an animated droppable placeholder. My thoughts:

- It might be better to make placeholder a connected component rather than passing down the phase via props.
- The way I've done dynamic CSS animations feels a bit 😕. I'm not sure if CSS transitions will work cleanly but it might be worth exploring. Otherwise we might have to do animations with JS.
- There's probably a layer of memoization and optimisation which could sit on top of some of the functions in here.

Anyway, proof of concept 🤗


alexreardon:
As discussed - we will not be tackling this as part of #2 in order to ship the feature a bit quicker. We can add this in afterwards 👍 

Issue to track this word: #77

alexreardon:
For future purposes: I think it will be cleaner to avoid needing to manage a stylesheet for the animation. We can just use transition and some componentDidMount trickery. Similar to what we do with [Reveal in @atlaskit/navigation](https://aui-cdn.atlassian.com/atlaskit/stories/@atlaskit/navigation/21.2.2/?selectedKind=%40atlaskit%2Fnavigation%20-%20Reveal%20%28interal%29&selectedStory=mount%20open%20with%20no%20animation&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel)

PR #76, opened 2017-09-06
leaning on eslint config rather than options to control linting targets

alexreardon:


alexreardon:
@lukebatchelor do you think this is the right way to go? I think so...

PR #80, opened 2017-09-08
Jaredcrowe/moving lists disabled droppables

jaredcrowe:


src/state/get-drag-impact.js
```diff
@@ -77,6 +77,22 @@ export default ({
 
   const axis: Axis = droppable.axis;
 
+  if (!droppable.isEnabled) {
+    const homeDroppableId = draggingDimension.droppableId;
+    const homeIndex = getDraggablesInsideDroppable(
+      droppables[homeDroppableId],
+      draggables,
+    ).indexOf(draggingDimension);
+    return {
```
jaredcrowe: We need to actually return an impact with a destination or else we get errors when trying to lift from a disabled droppable.

stories/src/multiple-vertical/quote-app.jsx
```diff
@@ -121,8 +128,19 @@ export default class QuoteApp extends Component {
     `;
   }
 
+  getDisabledDroppable = (sourceDroppable: ?string) => {
```
jaredcrowe: I updated the stress test story to demo this - when you lift from a droppable the next droppable will dynamically become disabled.

src/state/get-drag-impact.js
alexreardon: perfect

src/state/action-creators.js
```diff
@@ -130,6 +130,23 @@ export const updateDroppableDimensionScroll =
     },
   });
 
+export type UpdateDroppableIsEnabledAction = {|
```
alexreardon: love it

src/state/get-drag-impact.js
alexreardon: can you please add that as a comment in the code?

src/state/reducer.js
```diff
@@ -339,6 +339,23 @@ export default (state: State = clean('IDLE'), action: Action): State => {
...
 
+  if (action.type === 'UPDATE_DROPPABLE_IS_ENABLED') {
+    const { id, isEnabled } = action.payload;
+    return {
+      ...state,
+      dimension: {
+        ...state.dimension,
+        droppable: {
+          ...state.dimension.droppable,
+          [id]: {
```
alexreardon: rather than going this deep can we pull the dimension out?
Also, some error checking would be good also

```js
const target = state.dimension.droppable[id]

if(!target) {
  console.error('cannot update enabled flag on droppable that does not have a dimension');
  return clean();
}

if(target.isEnabled === isEnabled) {
  console.warn(`trying to set droppable isEnabled to ${isEnabled}` but it is already ${isEnabled}`);
  return state;
}

const updated = {
 ...target,
 isEnabled,
};

return {
       ...state,
       dimension: {
         ...state.dimension,
         droppable: {
           ...state.dimension.droppable,
           [id]: updated,
      }
 }
```

alexreardon:
What you have looks great - i have just suggested some changes - let me know when you want me to take another look

stories/src/multiple-vertical/quote-app.jsx
alexreardon: I was thinking of creating a little button where we can set the disabled / enabled status of a droppable in the story - don't need to do it here though. Thoughts?

alexreardon: we can leave this for now though - we can build on it

jaredcrowe: yeah that's something we can add later when we polish the stories IMO - this is just a quick and dirty way to test it now :)

alexreardon:
once the conflict is dealt will - shipit :)

PR #81, opened 2017-09-08
Add support for a rollup build

treshugart:
- [ ] Fix errors with `prop-types`, `react` and `reselect` dependencies.
- [ ] Ensure only required `dependencies` are being included in the bundle.
- [ ] Ensure all externals (i.e. `react`) have the correct global specified for the UMD global.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ node_modules/
 
 # generated files
 lib/
+umd/
 
 # generated site
 site/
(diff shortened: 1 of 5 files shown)


alexreardon:
Awesome @treshugart ! I was not expecting a PR - but it is most welcome!

treshugart:
I expected it to be easier, but lots of problems with common js libs.

alexreardon:
As discussed IRL - for now I am hoping to not maintain a UMD build. My original thinking was around whether we should pre-optimise the lib before bundler consumption.

Thanks for jumping on this @treshugart

PR #82, opened 2017-09-08
add tests for dragging between lists to get-drag-impact.spec

jaredcrowe:


test/utils/create-droppable.js
```diff
@@ -0,0 +1,69 @@
...
+
+type TestDroppable = {
+  droppableId: string,
+  droppable: DroppableDimension,
+  draggables: DraggableDimensionMap,
+  draggableIds: string[],
+  draggableDimensions: DraggableDimension[],
+};
+
+export default ({
```
jaredcrowe: ended up moving this out as a util because I find it super handy

test/unit/state/get-drag-impact.spec.js
```diff
@@ -1133,4 +1135,213 @@ describe('get drag impact', () => {
       });
     });
   });
+
+  describe('moving between lists', () => {
+    const homeDroppable = createDroppable({
```
alexreardon: love this

test/utils/create-droppable.js
alexreardon: this is great!

PR #85, opened 2017-09-08
Decrease minimal react version ( closed #84 )

### LINKED ISSUE #84: Problem with strict peerDependency
Firstly, thanks for this component, it's great.

Why this component has strict peerDependency react `15.6.1`?
Have you real causes for it?
Could it be changed to somthing like `>=15.4.0`?
We cannot use your component if we have other react version.
We very disappointed about it.

Thanks.

pofigizm:
All cli tests passed.
Hand tests is ok.
- Chrome, Firefox, Safari (mac)
- IE 11 (win7 in VirtualBox)

I did't change `react-dom` version, because `react-dom/test-utils` used in tests (since 15.5)
But I made hand tests with `react-dom 15.4.0`

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "flow-copy-source": "1.2.1",
     "jest": "20.0.4",
     "raf-stub": "2.0.1",
-    "react": "15.4.0",
+    "react": "15.6.1",
     "react-dom": "15.6.1",
     "react-test-renderer": "15.6.1",
     "styled-components": "2.1.2"
--- a/package.json
+++ b/package.json
@@ -64,13 +64,13 @@
     "flow-copy-source": "1.2.1",
     "jest": "20.0.4",
     "raf-stub": "2.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -64,13 +64,13 @@
     "flow-copy-source": "1.2.1",
     "jest": "20.0.4",
     "raf-stub": "2.0.1",
-    "react": "15.6.1",
```
alexreardon: Can we please keep the dev dependency version the same and just change the peerDependency version?

pofigizm: It's good practice to use same version in develop. So you can quickly find breaking changes. But, it isn't problem for me. I've done it.

pofigizm:
Hi, @alexreardon.
Thank you for approve.
Are you planning to merge it? How soon will it be released?

alexreardon:
I am deep in work getting #2 finished so i have a bit of tunnel vision. I will be doing a release soon

PR #88, opened 2017-09-14
droppable placeholder now takes up correct size

alexreardon:
Without this the droppable placeholder is too big


test/unit/view/connected-droppable.spec.js
```diff
@@ -60,9 +61,14 @@ type DragArgs = {|
   isDraggingOver: false | 'home' | 'foreign'
 |}
 
+const margin: Spacing = {
```
alexreardon: the tests where asserting the correct values - but because there was no margin here `withMargin` was the same as `withoutMargin`

PR #92, opened 2017-09-15
get drag impact refactor

alexreardon:
This PR is designed to simplify how we handle `page` values in the system - as well as the `get-drag-impact` function. I noticed that the `get-drag-impact` function was more complicated then it needed to be so I have simplified how it works. 

This pull request does involve a functional change: currently the impact of a drag is based on the mouse selection point and not the center point which was not intended. Fixing that is as simple as changing the argument to getDragImpact from page.selection to page.center. However, I have gone all out and decided to clean the whole thing up a big before we move on.

I am providing all this detail in case i am unavailable to continue with it @jaredcrowe.


src/state/get-drag-impact.js
```diff
@@ -38,63 +37,57 @@ const getDroppablesScrollDiff = ({
...
-      direction: axis.direction,
-      destination: null,
-    };
+  if (!destination.isEnabled) {
+    return noImpact;
   }
 
-  const insideDroppable: DraggableDimension[] = getDraggablesInsideDroppable(
-    droppable,
+  const source: DroppableDimension = droppables[draggable.droppableId];
```
alexreardon: hopefully this is much clearer now

src/state/reducer.js
```diff
@@ -79,55 +77,38 @@ const move = ({
 
   const previous: CurrentDrag = state.drag.current;
   const initial: InitialDrag = state.drag.initial;
-  const droppable: DroppableDimension = state.dimension.droppable[initial.source.droppableId];
+  const currentWindowScroll: Position = windowScroll || previous.windowScroll;
```
alexreardon: rather than everywhere calculating page - it does just done here based on the window scroll

src/state/reducer.js
```diff
@@ -79,55 +77,38 @@ const move = ({
...
-      center,
+      selection: clientSelection,
+      center: add(offset, initial.client.center),
     };
     return result;
   })();
 
-  const scrollDiff: Position = subtract(droppable.scroll.initial, droppable.scroll.current);
-
-  const withinDroppable: WithinDroppable = {
```
alexreardon: within droppable is gone. It was only used for get-drag-impact and the new function is much clearer

src/state/action-creators.js
```diff
@@ -69,7 +69,6 @@ export type CompleteLiftAction = {|
     id: DraggableId,
     type: TypeId,
     client: InitialDragLocation,
-    page: InitialDragLocation,
```
alexreardon: Removed page from all these actions - easy to compute once with windowScroll in the reducer

src/state/get-drag-impact.js
```diff
@@ -1,182 +0,0 @@
-// @flow
```
alexreardon: now split into different files for clarity and simplification

src/state/get-drag-impact/index.js
```diff
@@ -0,0 +1,68 @@
...
+type Args = {|
+  pageCenter: Position,
+  draggable: DraggableDimension,
+  // all dimensions in system
+  draggables: DraggableDimensionMap,
+  droppables: DroppableDimensionMap
+|}
+
+export default ({
+  pageCenter,
```
alexreardon: now calling this out explicitly

src/state/get-drag-impact/in-home-list.js
```diff
@@ -0,0 +1,100 @@
...
+
+export default ({
+  pageCenter,
+  draggable,
+  home,
+  insideHome,
+}: Args): DragImpact => {
+  const axis: Axis = home.axis;
+  const homeScrollDiff: Position = subtract(home.scroll.current, home.scroll.initial);
+  const originalCenter: Position = draggable.page.withoutMargin.center;
```
alexreardon: TODO: check this position

alexreardon: This is correct - the ordering of the variables was a little funny so i have cleaned that up

PR #95, opened 2017-09-18
board story demonstrates issue with margin collapse + placeholder

jaredcrowe:
just creating this PR to get the storybook build, feel free to decline it

### DIFF
--- a/stories/src/primatives/quote-list.jsx
+++ b/stories/src/primatives/quote-list.jsx
@@ -18,7 +18,7 @@ const Wrapper = styled.div`
   display: flex;
   flex-direction: column;
   opacity: ${({ isDropDisabled }) => (isDropDisabled ? 0.5 : 'inherit')};
-  padding: ${grid}px;
+  padding: 0 ${grid}px;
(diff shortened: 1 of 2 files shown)


stories/src/primatives/quote-list.jsx
```diff
@@ -18,7 +18,7 @@ const Wrapper = styled.div`
   display: flex;
   flex-direction: column;
   opacity: ${({ isDropDisabled }) => (isDropDisabled ? 0.5 : 'inherit')};
-  padding: ${grid}px;
```
alexreardon: I am not sure if it is feasible to detect for margin collapsing for droppable placeholders. My current thinking is to add a note about it in the docs and leave it at that. Thoughts?

jaredcrowe: might have to be. I'm spiking a solution atm but I'm not sure if it'll cover everything. I'll report back 👨‍🚀

alexreardon:
Can we close this one?

jaredcrowe:
yup 👍

PR #98, opened 2017-09-19
fix droppable dimension calculation when within a shared scroll conta…

jaredcrowe:
…iner

### DIFF
--- a/src/view/droppable-dimension-publisher/droppable-dimension-publisher.jsx
+++ b/src/view/droppable-dimension-publisher/droppable-dimension-publisher.jsx
@@ -72,16 +72,9 @@ export default class DroppableDimensionPublisher extends Component {
         return current;
       }
 
+      const { top, right, bottom, left } = current;
+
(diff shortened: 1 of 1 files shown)


src/view/droppable-dimension-publisher/droppable-dimension-publisher.jsx
```diff
@@ -72,16 +72,9 @@ export default class DroppableDimensionPublisher extends Component {
...
-      // We need to trim the dimension by the visible area of the scroll container
-
-      // Adjust the current dimension with the parents scroll
-      const top = current.top + scroll.y;
-      const bottom = current.bottom + scroll.y;
-      const left = current.left + scroll.x;
-      const right = current.right + scroll.x;
-
-      // Trim the dimension by the size of the parent
+      const { top, right, bottom, left } = current;
```
jaredcrowe: It seems to honestly be as simple as this, but I'm hesitant to believe it's this easy... 🤔

PR #99, opened 2017-09-19
fix dragging between droppables within a shared scroll container

jaredcrowe:


jaredcrowe:
I don't want to change the stories as part of my fix, so here are the stories to demo that the fix actually works. don't actually merge this :)

src/state/dimension.js
```diff
@@ -154,6 +153,14 @@ export const getDroppableDimension = ({
       withMargin: getFragment(getWithSpacing(withWindowScroll, margin)),
       withMarginAndPadding: getFragment(getWithSpacing(withWindowScroll, add(margin, padding))),
     },
+    container: {
+      scroll: {
```
alexreardon: In some cases the droppable is the scrollable element itself - are we catering for that? And does the language used reflect that?

src/state/is-within-visible-bounds-of-droppable.js
```diff
@@ -8,10 +9,32 @@ import type {
...
+  const droppableRect = droppable.page.withMarginAndPadding;
+  const { container } = droppable;
 
-  const isWithinHorizontal = isWithin(left, right);
-  const isWithinVertical = isWithin(top, bottom);
+  // Calculate the mid-drag scroll ∆ of the scroll container
+  const containerScrollDiff = subtract(container.scroll.current, container.scroll.initial);
+
+  // Calculate the droppable's bounds, accounting for the container's scroll
+  const droppableBounds = {
```
alexreardon: run time clipping - love it

alexreardon: we seem to do this type of operation a lot: `Spaceing` add or subtract a `position`. Perhaps it is worth creating a utlity for this in a `spacing.js` file

src/types.js
```diff
@@ -99,6 +95,13 @@ export type DroppableDimension = {|
     // the area in which content presses up against
     withMarginAndPadding: DimensionFragment,
   |},
+  container: {|
+    scroll: {|
+      initial: Position,
+      current: Position,
+    |},
+    page: DimensionFragment,
```
alexreardon: to follow conventions in the other fragments: should this be broken into `withMargin` and `withoutMargin`?

src/types.js
```diff
@@ -99,6 +95,13 @@ export type DroppableDimension = {|
     // the area in which content presses up against
     withMarginAndPadding: DimensionFragment,
   |},
+  container: {|
```
alexreardon: I am a little worried that it might not be clear that the 'container' might actually be itself. Perhaps the label `closestScrollable` is clearer.

alexreardon: An alternative type:

```js
export type DroppableDimension = {|
  id: DroppableId,
  axis: Axis,
  isEnabled: boolean,
  // relative to the current viewport
  client: {|
    withMargin: DimensionFragment,
    withoutMargin: DimensionFragment,
    // the area in which content presses up against
    withMarginAndPadding: DimensionFragment,
  |},
  // relative to the whole page
  page: {|
    withMargin: DimensionFragment,
    withoutMargin: DimensionFragment,
    // the area in which content presses up against
    withMarginAndPadding: DimensionFragment,
  |},
  closestScrollable: {|
      scroll: {|
        initial: Position,
        current: Position,
      |},
      page: {|
       withMargin: DimensionFragment,
       withoutMargin: DimensionFragment,
     |}
  |},
|}
```

alexreardon:
I am really excited that you got this working - well done @jaredcrowe

alexreardon:
It would be good to check it all also works with window scrolling

src/types.js
jaredcrowe: yeah it's a tough one because sometimes the droppable won't have to consider scrolling at all - it won't have internal scrolling and it won't be within a scroll parent so its container will be itself.

essentially what this dimension represents is the maximum possible extent of the droppable's visible bounds. this could be either:
* its own dimensions (when it has no internal scrolling and isn't inside a scrolling parent)
* its max-height/max-width (when it has internal scrolling)
* the dimension of its scroll parent (when it's inside a scrolling parent). note: in this case its container dimension may actually be larger than its own dimension (imagine multiple droppables inside a shared scroll parent), but this scroll parent still constrains its potential visible area.

in light of this, 'container' seemed like the most appropriately generic terminology... 🤔  do you have any other suggestions? totally open to changing this if something else makes more sense to you :)

jaredcrowe:
I've put together the mother of all stress tests and it works with everything I can possibly think of 😛 
https://deploy-preview-99--react-beautiful-dnd.netlify.com/?selectedKind=multiple%20vertical%20lists&selectedStory=stress%20test&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel
let me know if you can think of any scenario I've missed

src/types.js
jaredcrowe: more diagrams 🙌
container_dimensions

alexreardon: I'll leave it to you about name - but definitely add a good comment explaining what it is. The diagrams look great - be sure to include them in the release notes if it makes sense :)

What do you think about making `page` > `withMargin` and `withoutMargin` (for consistency)

jaredcrowe: 🤔 we can add a withoutMargin property for object shape consistency if you want to, but I don't think we'll ever care about a container's withMargin dimensions because we only need to know its content box.

I'll do something like this for consistency:
```javascript
container: {|
  scroll: {|
    initial: Position,
    current: Position,
  |},
  page: {|
    withoutMargin: DimensionFragment,
  |},
|}
```

alexreardon: perfect

jaredcrowe:
@alexreardon this is ready for a final review if you have a chance. Otherwise I'll merge tomorrow 🙂

src/state/is-within-visible-bounds-of-droppable.js
```diff
@@ -8,10 +9,26 @@ import type {
 } from '../types';
 
 export const isPointWithin = (droppable: DroppableDimension) => {
-  const { top, right, bottom, left } = droppable.page.withMargin;
+  // Calculate the mid-drag scroll ∆ of the scroll container
```
alexreardon: Can you add the types to each of these variables for clarity?

src/state/is-within-visible-bounds-of-droppable.js
```diff
@@ -8,10 +9,26 @@ import type {
...
-  const isWithinHorizontal = isWithin(left, right);
-  const isWithinVertical = isWithin(top, bottom);
+  // Calculate the droppable's bounds, accounting for the container's scroll
+  const droppableBounds = offsetSpacing(droppable.page.withMargin, containerScrollDiff);
+
+  // Clip the droppable's bounds by the scroll container's bounds
+  // This gives us the droppable's true visible area
+  // Note: if the droppable doesn't have a scroll parent droppableBounds === container.page
+  const containerBounds = droppable.container.page.withoutMargin;
+  const clippedBounds = {
```
alexreardon: FYI this is of type Spacing

src/state/position.js
```diff
@@ -51,3 +51,10 @@ export const distance = (point1: Position, point2: Position): number =>
 // When given a list of points, it finds the smallest distance to any point
 export const closest = (target: Position, points: Position[]): number =>
   Math.min(...points.map((point: Position) => distance(target, point)));
+
+export const offsetSpacing = (spacing: Spacing, position: Position) => ({
```
alexreardon: Can you annotate this to say it returns a type Spacing

test/unit/view/unconnected-droppable-dimension-publisher.spec.js
```diff
@@ -323,359 +321,133 @@ describe('DraggableDimensionPublisher', () => {
...
-
-        // the trimmed rect
-        expect(publish).toBeCalledWith(expected);
-        expect(publish).toHaveBeenCalledTimes(1);
+      it('should detect scrolling on the droppable', () => {
+        wrapper = mount(<App
+          droppableIsScrollable
+          parentIsScrollable
+          onPublish={(dimension) => {
+            expect(dimension).toBeTruthy();
```
alexreardon: I generally do not like doing assertions like this. Any chance this can call a mock and then you check what the mock was called with in the assertion phase?

src/state/get-drag-impact/in-home-list.js
```diff
@@ -28,7 +28,9 @@ export default ({
   insideHome,
 }: Args): DragImpact => {
   const axis: Axis = home.axis;
-  const homeScrollDiff: Position = subtract(home.scroll.current, home.scroll.initial);
+  const homeScrollDiff: Position = subtract(
```
alexreardon: Was the test case for this file updated?

jaredcrowe: Nothing changed here functionally, I just had to update it because the shape of `DroppableDimension` changed.

alexreardon: :thumbsup:

src/state/is-within-visible-bounds-of-droppable.js
```diff
@@ -1,17 +1,35 @@
...
   DraggableDimension,
   DroppableDimension,
   DimensionFragment,
+  Spacing,
 } from '../types';
 
 export const isPointWithin = (droppable: DroppableDimension) => {
-  const { top, right, bottom, left } = droppable.page.withMargin;
+  // Calculate the mid-drag scroll ∆ of the scroll container
+  const scroll = droppable.container.scroll;
```
alexreardon: scroll: Position

src/state/is-within-visible-bounds-of-droppable.js
```diff
@@ -1,17 +1,35 @@
...
-  const isWithinHorizontal = isWithin(left, right);
-  const isWithinVertical = isWithin(top, bottom);
+  // Calculate the droppable's bounds, accounting for the container's scroll
+  const droppableBounds: Spacing = offsetSpacing(droppable.page.withMargin, containerScrollDiff);
+
+  // Clip the droppable's bounds by the scroll container's bounds
+  // This gives us the droppable's true visible area
+  // Note: if the droppable doesn't have a scroll parent droppableBounds === container.page
+  const containerBounds = droppable.container.page.withoutMargin;
+  const clippedBounds: Spacing = {
```
alexreardon: nice

src/state/is-within-visible-bounds-of-droppable.js
jaredcrowe: scroll actually has this shape:
```javascript
{
  current: Position,
  initial: Position,
}
```
I think it's OK for this not to be annotated because there are no moving parts here - I'm just pulling the prop off the `droppable` to make the code more readable. We're already checking the type of `droppable` when it comes in so there's no chance of anything going wrong here. Same goes for `containerBounds` below.

PR #101, opened 2017-09-19
Fixing nested keyboard dragging

alexreardon:


alexreardon:
daymn, look at all those commits - i'll try to clean that up

PR #102, opened 2017-09-19
fixing keyboard dropping when inside of a nested list

alexreardon:


test/unit/view/drag-handle.spec.js
```diff
@@ -1005,6 +1005,18 @@ describe('drag handle', () => {
           onKeyLift: 0,
         })).toBe(true);
       });
+
+      it('should not lift if disabled', () => {
```
alexreardon: lifting restriction where spacebar could not be used during a mouse drag - now that we listen to the window scroll we no longer need this! 🎉

PR #104, opened 2017-09-20
adding movement description to docs

alexreardon:


alexreardon:
After this PR I will be releasing a new version

PR #113, opened 2017-09-24
Flow cleanup

alexreardon:
This PR fixes a few issues:

- adding `babel-plugin-react-flow-props-to-prop-types` as a production dependency (because it is one)
- adding `rimraf` to safely remove `lib` folder before it is build. This prevents previously generated `/lib` files that are no longer needed from being accidentally published to npm
- no longer using fuzzy `redux` version.

I have tested this out here: https://github.com/alexreardon/react-beautiful-dnd-flow-example


alexreardon:
I am keen to remove the `prop-type` conversion and just lean on flow exclusively. However, I thought it better to get it working first and then move from there

alexreardon:
I will fix these conflicts soon

rhys-vdw:
Hey @alexreardon FYI this broke our build in IE.

> adding babel-plugin-react-flow-props-to-prop-types as a production dependency (because it is one)

I don't know exactly what that package is (babel plugin doesn't *sound* like a prod dependency). Regardless it depends on `chalk`, which apparently does not transpile it's code for browser compatibility (includes usage of `const`).

```
-> yarn why chalk
yarn why v1.3.2
warning ../../../package.json: No license field
[1/4] 🤔  Why do we have the module "chalk"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
info Has been hoisted to "chalk"
info Reasons this module exists
   - "react-beautiful-dnd#babel-plugin-react-flow-props-to-prop-types#babel-file-loader#babel-file#babel-core#babel-code-frame" depends on it
```

By listing it as a dependency we were unable to run our app in IE11 in development do to a "const is not defined" error in IE11. I think this only affected us in dev because we were using `DLLPlugin` - I don't have time to look into it further.

No biggy, but I thought I might as well let you know since it took me a while to work out what was going on.

alexreardon:
Interesting! This should not be a current issue as the `babel-plugin-react-flow-props-to-prop-types` was removed a long time ago. What version of `react-beautiful-dnd` are you running?

rhys-vdw:
Yep, I'm aware, we can easily sidestep this. We're on 3.0.0 but I'll upgrade.

PR #123, opened 2017-09-27
dimension clipping now considers margins

### LINKED ISSUE #122: Dragging to Bottom of a "Full" Foreign List Not Working
## Bug or feature request?
Bug - Can't drag to a new list from the bottom when the list is "full" (no extra droppable space)

### Expected behaviour
Should be able to drop to a list from below, and a placeholder should appear.

### Actual behaviour
Can't drop to a new list from its end.

### Steps to reproduce
Please see attached gif.

### Browser version
Version 60.0.3112.113 (Official Build) (64-bit)

### Demo
dragbelow

alexreardon:
Fixes #122. 

@jaredcrowe can you please verify that this is what we want to be doing. I think it is. I had a play with the storybooks and and it yields the expected results. 

This PR is not quite done - I have not finished fixing all of the tests in `is-within-visible-bounds-of-droppable.spec`. Can you please update the remaining ones?

Cheers

### DIFF
--- a/src/state/dimension.js
+++ b/src/state/dimension.js
@@ -1,6 +1,7 @@
 // @flow
 import { vertical, horizontal } from './axis';
 import getClientRect from './get-client-rect';
+import { add, isEqual } from './spacing';
 import type {
   DroppableId,
   DraggableId,
@@ -116,20 +117,6 @@ type GetDroppableArgs = {|
   isEnabled?: boolean,
 |}
 
-const add = (spacing1: Spacing, spacing2: Spacing): Spacing => ({
-  top: spacing1.top + spacing2.top,
-  left: spacing1.left + spacing2.left,
-  right: spacing1.right + spacing2.right,
-  bottom: spacing1.bottom + spacing2.bottom,
-});
-
-const equal = (spacing1: Spacing, spacing2: Spacing): boolean => (
-  spacing1.top === spacing2.top &&
-  spacing1.right === spacing2.right &&
-  spacing1.bottom === spacing2.bottom &&
-  spacing1.left === spacing2.left
-);
-
 export const getDroppableDimension = ({
   id,
   clientRect,
@@ -144,11 +131,12 @@ export const getDroppableDimension = ({
   const withMargin = getWithSpacing(clientRect, margin);
   const withWindowScroll = getWithPosition(clientRect, windowScroll);
   // If no containerRect is provided, or if the clientRect matches the containerRect, this
-  // droppable is its own container. In that case we include margin in the container dimension.
-  const droppableIsContainer = !containerRect || equal(containerRect, clientRect);
-  const containerRectWithWindowScroll = droppableIsContainer
+  // droppable is its own container. In this case we include its margin in the container bounds.
+  // Otherwise, the container is a scrollable parent. In this case we don't care about margins
+  // in the container bounds.
+  const containerRectWithWindowScroll = !containerRect || isEqual(containerRect, clientRect)
     ? getWithPosition(withMargin, windowScroll)
-    : getWithPosition(containerRect || withMargin, windowScroll);
+    : getWithPosition(containerRect, windowScroll);
 
   const dimension: DroppableDimension = {
     id,
@@ -170,9 +158,7 @@ export const getDroppableDimension = ({
         // when we start the current scroll is the initial scroll
         current: scroll,
       },
-      page: {
-        withoutMargin: getFragment(containerRectWithWindowScroll),
-      },
+      bounds: getFragment(containerRectWithWindowScroll),
     },
   };
 
--- a/src/state/dimension.js
+++ b/src/state/dimension.js
@@ -123,6 +123,13 @@ const add = (spacing1: Spacing, spacing2: Spacing): Spacing => ({
   bottom: spacing1.bottom + spacing2.bottom,
 });
 
+const equal = (spacing1: Spacing, spacing2: Spacing): boolean => (
+  spacing1.top === spacing2.top &&
+  spacing1.right === spacing2.right &&
+  spacing1.bottom === spacing2.bottom &&
+  spacing1.left === spacing2.left
+);
+
 export const getDroppableDimension = ({
   id,
   clientRect,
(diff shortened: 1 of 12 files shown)


src/state/dimension.js
```diff
@@ -161,6 +161,7 @@ export const getDroppableDimension = ({
       },
       page: {
         withoutMargin: getFragment(containerRectWithWindowScroll),
+        withMargin: getFragment(getWithSpacing(containerRectWithWindowScroll, margin)),
```
alexreardon: This is the margin of the droppable and not the container - perhaps the container margin should be an extra argument?

jaredcrowe:
personally I don't know if including margin in the droppable's hitbox is the behaviour I'd expect. the real fix for #122 is to conditionally include the size of the placeholder in a droppable's dimensions if it's a foreign list and was hovered over on the previous tick. this is actually something I've got working in my auto-scrolling branch, so I might break it out into its own PR and try to merge it separately.

alexreardon:
I think I would expect the hit box to include margins as it is the visible area. Thoughts?

Regarding adding the placeholder, just be sure that it still works when lists are one after the other :) (no overlapping dimensions)

jaredcrowe:
Margins are for creating space around the outside of an element. i.e. I wouldn't consider the margin in this screenshot to be considered part of the droppable's hitbox. I'd only consider myself 'inside' the droppable if I was over the light blue area:
screen shot 2017-09-28 at 5 08 06 pm

If we're running into an issue with this then I think it's either:
a) we're attaching the droppable ref to the wrong element; or
b) we're using margins for spacing when we should be using padding

alexreardon:
I guess it depends how you use the margin. In our example we use margin for inner spacing. Also using margins would yield a bigger drop zone which is preferable over dead zones

jaredcrowe:
yeah I suppose that's true. OK I'll move ahead with this PR and get those tests working.

src/state/dimension.js
```diff
@@ -136,8 +143,12 @@ export const getDroppableDimension = ({
 }: GetDroppableArgs): DroppableDimension => {
   const withMargin = getWithSpacing(clientRect, margin);
   const withWindowScroll = getWithPosition(clientRect, windowScroll);
-  // If no containerRect is provided, the clientRect is the container
-  const containerRectWithWindowScroll = getWithPosition(containerRect || clientRect, windowScroll);
+  // If no containerRect is provided, or if the clientRect matches the containerRect, this
```
jaredcrowe: This is the new fix. I realised that we don't always want to track the container's margin - we only want to include a droppable's margin in the container dimension when the droppable is its own container.

test/unit/state/dimension.spec.js
```diff
@@ -313,5 +313,82 @@ describe('dimension', () => {
         expect(dimension.page.withMarginAndPadding).toEqual(fragment);
       });
     });
+
+    describe('calculating container dimension', () => {
```
jaredcrowe: added a bunch of tests to this suite to ensure that margin is considered when calculating container dimension correctly

src/state/dimension.js
```diff
@@ -136,8 +143,12 @@ export const getDroppableDimension = ({
...
+  // droppable is its own container. In that case we include margin in the container dimension.
+  const droppableIsContainer = !containerRect || equal(containerRect, clientRect);
+  const containerRectWithWindowScroll = droppableIsContainer
+    ? getWithPosition(withMargin, windowScroll)
+    : getWithPosition(containerRect || withMargin, windowScroll);
```
alexreardon: can we ever have not have a containerRect when the droppable is not the container?

src/state/dimension.js
```diff
@@ -136,8 +143,12 @@ export const getDroppableDimension = ({
 }: GetDroppableArgs): DroppableDimension => {
   const withMargin = getWithSpacing(clientRect, margin);
   const withWindowScroll = getWithPosition(clientRect, windowScroll);
-  // If no containerRect is provided, the clientRect is the container
-  const containerRectWithWindowScroll = getWithPosition(containerRect || clientRect, windowScroll);
+  // If no containerRect is provided, or if the clientRect matches the containerRect, this
+  // droppable is its own container. In that case we include margin in the container dimension.
+  const droppableIsContainer = !containerRect || equal(containerRect, clientRect);
+  const containerRectWithWindowScroll = droppableIsContainer
```
alexreardon: line 174 (page > withoutMargin) should be renamed now - sometimes it includes margins and sometimes it doesn't...

src/state/dimension.js
alexreardon: should we capture the margin of the container, rather than just its clientRect?

jaredcrowe: no, but flow was having a sad about it 😕 I might re-write the logic so flow realises that it's impossible for containerRect to be undefined at this point.

jaredcrowe: I think in a case where the container isn't the droppable (i.e. the second or third case in [this diagram](https://user-images.githubusercontent.com/4995433/30788964-4888a78a-a1e5-11e7-8268-14597f2b0f6d.png)) we wouldn't want to know about the container's margin. In the first case we take the droppable's margin into account.

alexreardon: Yeah, perhaps just add a line that throws if clientRect is null. It would be clearer then :)

alexreardon:
You good for this one to go?

alexreardon:
If so I’ll take another look :)

jaredcrowe:
yup :)

src/state/dimension.js
```diff
@@ -136,8 +130,13 @@ export const getDroppableDimension = ({
 }: GetDroppableArgs): DroppableDimension => {
   const withMargin = getWithSpacing(clientRect, margin);
   const withWindowScroll = getWithPosition(clientRect, windowScroll);
-  // If no containerRect is provided, the clientRect is the container
-  const containerRectWithWindowScroll = getWithPosition(containerRect || clientRect, windowScroll);
+  // If no containerRect is provided, or if the clientRect matches the containerRect, this
+  // droppable is its own container. In this case we include its margin in the container bounds.
+  // Otherwise, the container is a scrollable parent. In this case we don't care about margins
+  // in the container bounds.
+  const containerRectWithWindowScroll = !containerRect || isEqual(containerRect, clientRect)
```
alexreardon: please add type: `:ClientRect`

src/state/spacing.js
```diff
@@ -0,0 +1,22 @@
+import type { Position, Spacing } from '../types';
+
+export const add = (spacing1: Spacing, spacing2: Spacing): Spacing => ({
```
alexreardon: at some point it will be good to add some tests for these functions, but for now it looks good

jaredcrowe: i did :) https://github.com/atlassian/react-beautiful-dnd/blob/b53e27efd85ae746c351805c1e98d43438d0f1c1/test/unit/state/spacing.spec.js

alexreardon: nice

PR #124, opened 2017-09-28
fix horizontal list story

jaredcrowe:
The issue arose because because in the horizontal plane the width of elements within a scroll container doesn't extend beyond their parent, even if they should be wider.


stories/src/primatives/author-list.jsx
```diff
@@ -20,8 +20,6 @@ const Wrapper = styled.div`
   user-select: none;
   transition: background-color 0.1s ease;
   margin: ${grid}px 0;
-
-  overflow: auto;
```
jaredcrowe: We had two levels of scroll containers in our horizontal list stories. Removing this to reduce confusion.

PR #125, opened 2017-09-28
fix race condition when cleaning state when a drop occurs before a li…

### LINKED ISSUE #121: "cannot drop if not dragging" (+ page stuck) when DnD quickly
## Bug or feature request?
Bug

### Expected behaviour
Drag and Drop will work (or even cancel) for quick drag and drop

### Actual behaviour
Dragged element is floating over the entire page (position: fixed) and prevent all click events in page.
This error is written in the console:
"cannot drop if not dragging"
image

### Steps to reproduce
Drag and drop quickly several time, you won't need more then 10 seconds.

### Browser version
Google chrome 61.0.3163.91 (Official Build) (64-bit)

### Demo

dnd bug
Code:

```javascript
const reorder = (list, startIndex, endIndex) => {
    const result = Array.from(list);
    const [removed] = result.splice(startIndex, 1);
    result.splice(endIndex, 0, removed);

    return result;
};

const grid = 8;

const getListStyle = (isDraggingOver) => ({
    background: isDraggingOver ? '#efefef' : ''
});

const getItemStyle = (draggableStyle, isDragging) => ({
    // some basic styles to make the items look a bit nicer
    // userSelect: isDragging ? 'none' : '',
    padding: grid * 2,
    margin: `0 0 ${grid}px 0`,
    cursor: 'move',

    // change background colour if dragging
    background: isDragging ? '' : '',

    // styles we need to apply on draggables
    ...draggableStyle
});

@observer
export default class CreateMapping extends React.Component {

<-- SOME COMPONENT FUNCTIONS -->
onDragEnd = (result) => {
        // dropped outside the list
        if (!result.destination) {
            return;
        }

        const items = reorder(
            TaskStore.taskData.input.mapping.rules,
            result.source.index,
            result.destination.index
        );
        TaskStore.taskData.input.mapping.rules = items;
    };

    render() {
        let rules = TaskStore.taskData.input.mapping ? TaskStore.taskData.input.mapping.rules : [];
        return (
            <div className="create-mapping">
                <MappingHeaders/>
                <DragDropContext onDragEnd={this.onDragEnd}>
                    <Droppable droppableId="droppable">
                        {(provided, snapshot) => (
                            <div ref={provided.innerRef} style={getListStyle(snapshot.isDraggingOver)}>
                                {rules.map((rule, i) => {
                                    return <Draggable key={i} draggableId={i}>
                                        {(provided, snapshot) => (
                                            <div>
                                                <div ref={provided.innerRef} style={getItemStyle(provided.draggableStyle, snapshot.isDragging)} {...provided.dragHandleProps}>
                                                    <div>
                                                    <MappingRow counter={this.counter} index={i} remove={this.removeDataset} setInitCounter={this.setInitCounter}/>
                                                    </div>
                                                </div>
                                            </div>
                                        )}
                                    </Draggable>
                                })}
                            </div>
                        )}
                    </Droppable>
                </DragDropContext>
                <div className="btn btn-primary" title="Add another mapping to view" onClick={this.addRule}>+</div>
            </div>

        )
    }
```

jaredcrowe:
…ft has completed

fixes #121

### DIFF
--- a/src/state/action-creators.js
+++ b/src/state/action-creators.js
@@ -158,13 +158,13 @@ export type MoveAction = {|
 export const move = (id: DraggableId,
   client: Position,
   windowScroll: Position): MoveAction => ({
-  type: 'MOVE',
-  payload: {
-    id,
-    client,
-    windowScroll,
-  },
-});
+    type: 'MOVE',
+    payload: {
+      id,
+      client,
+      windowScroll,
+    },
+  });
 
 export type MoveByWindowScrollAction = {|
   type: 'MOVE_BY_WINDOW_SCROLL',
@@ -291,7 +291,9 @@ export const drop = () =>
 
     if (state.phase !== 'DRAGGING') {
       console.error('cannot drop if not dragging', state);
-      dispatch(clean());
+      // We need to wrap this in a setTimeout to avoid a race
+      // condition with `lift`, which includes timeouts
+      setTimeout(() => dispatch(clean()));
       return;
     }
 
@@ -450,7 +452,7 @@ export const lift = (id: DraggableId,
   setTimeout(() => {
     const state: State = getState();
 
-    if (state.phase !== 'IDLE' || state.phase !== 'DRAG_COMPLETE') {
+    if (state.phase !== 'IDLE' && state.phase !== 'DRAG_COMPLETE') {
       dispatch(clean());
     }
 
--- a/src/state/action-creators.js
+++ b/src/state/action-creators.js
@@ -158,13 +158,13 @@ export type MoveAction = {|
 export const move = (id: DraggableId,
   client: Position,
   windowScroll: Position): MoveAction => ({
-    type: 'MOVE',
-    payload: {
-      id,
-      client,
-      windowScroll,
-    },
-  });
+  type: 'MOVE',
+  payload: {
+    id,
+    client,
+    windowScroll,
+  },
+});
 
 export type MoveByWindowScrollAction = {|
   type: 'MOVE_BY_WINDOW_SCROLL',


src/state/action-creators.js
```diff
@@ -158,13 +158,13 @@ export type MoveAction = {|
...
   client: Position,
   windowScroll: Position): MoveAction => ({
-  type: 'MOVE',
-  payload: {
-    id,
-    client,
-    windowScroll,
-  },
-});
+    type: 'MOVE',
```
jaredcrowe: eslint was complaining about the indenting here

src/state/action-creators.js
```diff
@@ -450,7 +452,7 @@ export const lift = (id: DraggableId,
   setTimeout(() => {
     const state: State = getState();
 
-    if (state.phase !== 'IDLE' || state.phase !== 'DRAG_COMPLETE') {
+    if (state.phase !== 'IDLE' && state.phase !== 'DRAG_COMPLETE') {
```
jaredcrowe: changed this because I assume this was the intended functionality? the original statement will always evaluate to true.

alexreardon: Can you explain the thinking around this one a little more?

alexreardon: I think it makes sense, but good to walk through

alexreardon: Was this line the bug?

src/state/action-creators.js
```diff
@@ -291,7 +291,9 @@ export const drop = () =>
 
     if (state.phase !== 'DRAGGING') {
       console.error('cannot drop if not dragging', state);
-      dispatch(clean());
+      // We need to wrap this in a setTimeout to avoid a race
+      // condition with `lift`, which includes timeouts
+      setTimeout(() => dispatch(clean()));
```
alexreardon: This fixes the freeze i think

jaredcrowe: This is the actual fix. The bug was occurring when a drop was triggered before a lift was completed. The `lift` action includes two `setTimeout` calls and ends up putting the app in the `DRAGGING` state. If a drop is triggered before the timeout callback firest the app ends up stuck in the `DRAGGING` state even though the item has already been dropped. The simplest way to fix this is to wrap the `clean()` action here in a timeout of its own. This means it will fire after the `lift` finishes and will ensure that the app is always put back in the `IDLE` state.

alexreardon: nice

src/state/action-creators.js
jaredcrowe: nah this was just something I noticed while I was debugging. `state.phase !== 'IDLE' || state.phase !== 'DRAG_COMPLETE'` will always evaluate to `true`.

src/state/action-creators.js
alexreardon: Seems reasonable for now. Do you think there is any additional checks we should do here?

jaredcrowe: the most robust thing to do would be to store a reference to the `setTimeout` somewhere so that we can clear it, but I think this is a much less complex solution. fwiw we won't even need that timeout if this issue gets resolved: https://github.com/chenglou/react-motion/issues/437

PR #127, opened 2017-09-30
Allowing dynamic hooks

### LINKED ISSUE #49: Dynamically change hooks after first render
I was playing with this library on a toy side-project, and I noticed I cannot dynamically change hooks, especially after the first render.

The gist of the code I had is as follows:

```js
import { DragDropContext, Droppable, Draggable } from "react-beautiful-dnd";

import { compose, withState } from "recompose";

// fake data generator
const getItems = count =>
    Array.from({ length: count }, (v, k) => k).map(k => ({
        id: `item-${k}`,
        content: `item ${k}`,
        type: generateType(k)
    }));

// ...

const onDragEnd = setCards => result => {
    // dropped outside the list
    if (!result.destination) {
        return;
    }

    setCards(prevCards => {
        const nextCards = reorder(
            prevCards,
            result.source.index,
            result.destination.index
        );
        return nextCards;
    });
};

const Board = props => {
    const { cards, setCards } = props;

    return (
        // NOTE: unable to replace onDragEnd after first render
        <DragDropContext onDragEnd={onDragEnd(setCards)}>
            <Droppable droppableId="droppable">
                {generateBoard(cards)}
            </Droppable>
        </DragDropContext>
    );
};

const cardsState = withState("cards", "setCards", getItems(10));

export default compose(cardsState)(Board);
```

Even though the drag and drop (which is awesome by the way) appears to work fine, after the first render, the generated function from `onDragEnd(setCards)` is not replaced within `DragDropContext`. 

This is likely to do with the fact there is no `componentWillReceiveProps()` (or similar) at https://github.com/atlassian/react-beautiful-dnd/blob/0f557e036fc5bdedeb5ba13de9540de28883aa4a/src/view/drag-drop-context/drag-drop-context.jsx#L17-L46

------

Even though the above example uses `recompose`, one may refactor the above to use `redux` and friends (e.g. `react-redux`), whereby the developer can pass `dispatch` (or a binded `dispatch`) to `onDragEnd()`. You cannot really assume `dispatch` won't change after the first render.

alexreardon:
Fixes #49


jaredcrowe:
are we testing the dynamically changing hooks functionality?

alexreardon:
Originally I had some but I was unhappy  with them. However I just thought of a better way. I’ll add some soon

alexreardon:
Would you be able to take another look @jaredcrowe?

PR #132, opened 2017-10-03
droppable dimensions include the size of the draggable placeholder if…

jaredcrowe:
… they are a foreign list and were hovered on the previous tick

This addresses #122 by adding the placeholder size to a droppable's dimensions when it's being hovered.

### DIFF
--- a/src/state/get-droppable-over.js
+++ b/src/state/get-droppable-over.js
@@ -1,17 +1,22 @@
 // @flow
 import { isPointWithinDroppable } from './is-within-visible-bounds-of-droppable';
 import { patch } from './position';
+import getDraggablesInsideDroppable from './get-draggables-inside-droppable';
 import type {
   DroppableId,
   Position,
   DraggableDimension,
+  DraggableDimensionMap,
   DroppableDimensionMap,
   DroppableDimension,
 } from '../types';
 
+const noPadding: Position = { x: 0, y: 0 };
+
 type Args = {
   target: Position,
   draggable: DraggableDimension,
+  draggables: DraggableDimensionMap,
   droppables: DroppableDimensionMap,
   previousDroppableOver: ?DroppableId,
 };
@@ -19,6 +24,7 @@ type Args = {
 export default ({
   target,
   draggable,
+  draggables,
   droppables,
   previousDroppableOver,
 }: Args): ?DroppableId => {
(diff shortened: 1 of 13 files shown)


jaredcrowe:
I've already noticed a bug with this - if a droppable has a min-height but not enough items to fill it out the placeholder size will still be added to its dimensions unnecessarily. I'll keep working on this.

alexreardon:
Can you please explain the reasoning behind this change? If we merge #123 is this needed?

src/state/get-droppable-over.js
```diff
@@ -1,22 +1,65 @@
...
   DroppableId,
   Position,
+  DraggableDimension,
+  DraggableDimensionMap,
   DroppableDimensionMap,
   DroppableDimension,
 } from '../types';
 
-export default (
+const noPadding: Position = { x: 0, y: 0 };
```
alexreardon: rather than using the name padding which is already a thing, let's use another name. What about 'buffer'? or 'extra'?

`noBuffer`

src/state/get-droppable-over.js
```diff
@@ -1,22 +1,65 @@
...
+  previousDroppableOver,
+}: Args): ?DroppableId => {
   const maybe: ?DroppableDimension =
     Object.keys(droppables)
       .map((id: DroppableId): DroppableDimension => droppables[id])
-      .find((droppable: DroppableDimension): boolean => (
-        isPointWithin(droppable)(target)
-      ));
+      .find((droppable: DroppableDimension): boolean => {
+        const placeholderPadding = (() => {
```
alexreardon: Can you add the flow type to this variable as well as to the others in this file?

`placeholderBuffer: Position`

src/state/is-within-visible-bounds-of-droppable.js
```diff
@@ -21,24 +26,37 @@ export const isPointWithin = (droppable: DroppableDimension) => {
   // This gives us the droppable's true visible area
   // Note: if the droppable doesn't have a scroll parent droppableBounds === container.page
   const containerBounds = droppable.container.page.withoutMargin;
-  const clippedBounds: Spacing = {
+  const clippedBounds = {
```
alexreardon: why loose the type?

src/state/is-within-visible-bounds-of-droppable.js
```diff
@@ -9,7 +9,12 @@ import type {
   Spacing,
 } from '../types';
 
-export const isPointWithin = (droppable: DroppableDimension) => {
+const noPadding: Position = { x: 0, y: 0 };
+
+const getVisibleBounds = (
```
alexreardon: rather than adding an additional parameter here, perhaps we should create a function that add a `Position` or a `Spacing` to a `DroppableDimension`

src/state/reducer.js
```diff
@@ -107,17 +107,25 @@ const move = ({
     windowScroll: currentWindowScroll,
   };
 
+  const previousDroppableOver = state.drag && state.drag.impact.destination
```
alexreardon: flow

src/state/reducer.js
```diff
@@ -107,17 +107,25 @@ const move = ({
...
     droppables: state.dimension.droppable,
+    previousDroppableOver,
   }));
 
   const drag: DragState = {
     initial,
     impact: newImpact,
     current,
+    previous: {
+      droppableOver: previousDroppableOver,
```
alexreardon: can this just be the `id` and we keep the dimension within the `state.dimension`?

alexreardon: Oh sorry, I just saw it was the Id. Can this be changed to `droppableOverId`

alexreardon:
Can you please also find another reviewer for this one?

alexreardon:
These algorithms are also used by keyboard dragging. We need to check that they also hold up for that. Those algorithms already have logic for moving into the last place of a foreign droppable

jaredcrowe:
hey @alexreardon so #123 makes it a bit easier to drag into the last position of a full droppable because it includes the margin in the hit-box. This PR gives us the functionality we really want - including a placeholder's size in a droppable's dimension when appropriate.

Something like this is going to be necessary for implementing auto-scrolling, but I decided to pull it out into its own PR and try to get it in sooner because it improves the experience RE #122 as well.

If you have a chance (and I totally understand if you don't!) I'd be keen to get your thoughts on this approach. Currently I'm calculating the droppable's dimensions in real time during the drag (in the same place where we perform the runtime clipping), but I'm also considering capturing a 'withPlaceholder' fragment while capturing dimensions. If it's possible to do it once there I think it might be more performant. Might spike this approach on a separate branch too... 🤔

alexreardon:
The upfront withPlaceholder sounds simplier to reason about just looking at it. Then everything to do with the dimension is in the one place

lukebatchelor:
Sweet! Obviously I dont have a the requisite knowledge to comprehensively review this, but I've sat down with Jared and gone through the problem and the solution. 

Happy to merge this and make any changes as necessary.

alexreardon:
I do not have time to review this properly so i'll lean on @lukebatchelor and @jaredcrowe. I will remove my review

PR #148, opened 2017-10-25
add ignoreContainerClipping prop

jaredcrowe:


jaredcrowe:
Currently this implementation still accounts for container scroll.

src/view/droppable-dimension-publisher/droppable-dimension-publisher-types.js
```diff
@@ -24,12 +24,13 @@ export type DispatchProps = {|
 |}
 
 export type OwnProps = {|
+  children?: ReactElement,
   droppableId: DroppableId,
   direction: Direction,
+  ignoreContainerClipping: boolean,
```
jaredcrowe: this is the only thing that's actually been added here.

src/view/droppable-dimension-publisher/droppable-dimension-publisher.jsx
```diff
@@ -64,11 +74,9 @@ export default class DroppableDimensionPublisher extends Component {
     const clientRect: ClientRect = targetRef.getBoundingClientRect();
 
     const containerRect: ClientRect = (() => {
-      if (!this.closestScrollable) {
-        return clientRect;
-      }
-
-      if (this.closestScrollable === targetRef) {
+      if (ignoreContainerClipping ||
```
jaredcrowe: this is the fix

alexreardon: the feature :)

src/view/droppable-dimension-publisher/droppable-dimension-publisher.jsx
```diff
@@ -23,6 +23,10 @@ export default class DroppableDimensionPublisher extends Component {
   /* eslint-disable react/sort-comp */
   props: Props;
 
+  static defaultProps = {
```
alexreardon: because this is an internal component we do not need to use default props - we can do the defaulting in the public `Droppable`

alexreardon:
Can you please add a docs as a part of this PR?

src/view/droppable/droppable-types.js
```diff
@@ -31,11 +31,12 @@ export type MapProps = {|
 |}
 
 export type OwnProps = {|
-  droppableId: DroppableId,
+  children: (Provided, StateSnapshot) => ?ReactElement,
   direction: HasDefaultProp<Direction>,
+  droppableId: DroppableId,
+  ignoreContainerClipping: boolean,
```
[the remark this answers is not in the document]
jaredcrowe: Is this just a convention so it's easier to reason about? Pretty sure it isn't necessary: https://flow.org/en/docs/react/components/#toc-using-default-props

src/view/droppable-dimension-publisher/droppable-dimension-publisher.jsx
```diff
@@ -63,19 +69,16 @@ export default class DroppableDimensionPublisher extends Component {
...
-      return getClientRect(
-        this.closestScrollable.getBoundingClientRect()
-      );
-    })();
+    // The droppable's own bounds should be treated as the
+    // container bounds in the following situations:
+    // 1. The consumer has opted in to ignoring container clipping
+    // 2. There is no scroll container
+    // 3. The droppable has internal scrolling
+    const containerRect: ClientRect = ignoreContainerClipping ||
```
jaredcrowe: this check here is the new feature 👌

PR #160, opened 2017-11-03
Remove disclaimer for moving between droppables

on3iro:
The feature has been implemented with v2.2.0 and therefore the
disclaimer is no longer necessary.


alexreardon:
Thanks!!!

on3iro:
Glad I could help :)

PR #168, opened 2017-11-09
Update README.md to match WebpackBin example

skroth:
The `getItemStyle` in README.md had a bug when I tried it but the one in the WebpackBin code worked perfectly. This commit brings the change in the WebpackBin example to README.md.


alexreardon:
Thanks @skroth ! A really valuable change!

PR #172, opened 2017-11-15
Improved drag flow

alexreardon:
Less `setTimeout` craziness - and a much more solid flow as a result!

### DIFF
--- a/src/state/action-creators.js
+++ b/src/state/action-creators.js
@@ -233,6 +233,16 @@ export const clean = (): CleanAction => ({
   payload: null,
 });
 
+type PrepareAction = {
+  type: 'PREPARE',
+  payload: null,
+}
+
+export const prepare = (): PrepareAction => ({
+  type: 'PREPARE',
+  payload: null,
+});
+
 export type DropAnimateAction = {
   type: 'DROP_ANIMATE',
   payload: {|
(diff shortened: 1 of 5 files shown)


src/state/action-creators.js
```diff
@@ -434,26 +439,28 @@ export const lift = (id: DraggableId,
...
   // need to allow a flush of react-motion
+  dispatch(prepare());
+
   setTimeout(() => {
     const state: State = getState();
 
-    if (state.phase !== 'IDLE' && state.phase !== 'DRAG_COMPLETE') {
-      dispatch(clean());
+    // drag cancelled before timeout finished
+    if (state.phase !== 'PREPARING') {
```
alexreardon: Previously we had no way of knowing if the drag was ended before this first timeout was fired. It is possible to do with a well orchestrated force press in touch dragging. By adding an extra phase we can know if something has changed in the elapsed time.

alexreardon:
To be a good citizen I also wrote some tests for the lift action-creator. Hopefully this will assist in preventing regressions in this super important part of the code

src/state/action-creators.js
```diff
@@ -55,14 +55,6 @@ export const requestDimensions = (type: TypeId): RequestDimensionsAction => ({
   payload: type,
 });
 
-export type BeginLiftAction = {|
```
alexreardon: no longer needed

test/unit/state/action-creators.spec.js
```diff
@@ -0,0 +1,249 @@
+// @flow
```
alexreardon: All new tests! 💃

src/state/action-creators.js
```diff
@@ -279,26 +281,21 @@ export const drop = () =>
...
       dispatch(clean());
       return;
     }
 
+    // dropped in another phase except for dragging - this is an error
     if (state.phase !== 'DRAGGING') {
-      console.error('cannot drop if not dragging', state);
-      // We need to wrap this in a setTimeout to avoid a race
-      // condition with `lift`, which includes timeouts
-      setTimeout(() => dispatch(clean()));
```
alexreardon: yuck. this is gone! 🎉

src/types.js
```diff
@@ -208,7 +208,30 @@ export type PendingDrop = {|
...
 
-export type Phase = 'IDLE' | 'COLLECTING_DIMENSIONS' | 'DRAGGING' | 'DROP_ANIMATING' | 'DROP_COMPLETE';
+export type Phase =
+  // The application rest state
+  'IDLE' |
+
+  // When a drag starts we need to flush any existing animations
+  // that might be occurring. While this flush is occurring we
+  // are in this phase
+  'PREPARING' |
```
alexreardon: new phase to allow for flushing

test/unit/state/action-creators.spec.js
```diff
@@ -0,0 +1,250 @@
...
+        const dispatch: Function = jest.fn();
+        const getState: Function = jest.fn(() => state);
+
+        liftWithDefaults()(dispatch, getState);
+
+        expect(dispatch).toHaveBeenCalledWith(clean());
+        expect(console.error).toHaveBeenCalled();
+      });
+
+      it('should not begin a lift if the drag is cancelled while the animations are flushing', () => {
```
alexreardon: This test executes a current error case - which is what caused me to make this change

alexreardon:
This will target master once #172 is merged

PR #175, opened 2017-11-15
fixing nested touch dragging

### LINKED ISSUE #173: Touch doesn't drag card but block in nested vertical list (Android only)
## Bug or feature request?
Bug

### Expected behavior
In this [example](https://react-beautiful-dnd.netlify.com/?selectedKind=complex%20vertical%20list&selectedStory=nested%20vertical%20lists&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel) in the `second level` block, the user should be able to drag and drop the cards using touch.

### Actual behavior
Using touch drags the whole `second level` block rather than just the card.

### Steps to reproduce
Go to [this doc page](https://react-beautiful-dnd.netlify.com/?selectedKind=complex%20vertical%20list&selectedStory=nested%20vertical%20lists&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel) with your phone (or Chrome in mobile mode) and try to swap the two cards in the `second level` block

### Browser version

Android: Chrome 62.0.3203.84
Computer: Chrome 62.0.3202.89 (Official Build) (64-bit)

alexreardon:
Fixes #173

### DIFF
--- a/src/view/drag-handle/sensor/create-touch-sensor.js
+++ b/src/view/drag-handle/sensor/create-touch-sensor.js
@@ -263,8 +263,12 @@ export default (callbacks: Callbacks, getDraggableRef: () => ?HTMLElement): Touc
       return;
     }
 
-    // Not stopping the event as we want to allow force press and other events to occur
-    // Scrolling is stopped by onTouchMove
+    // We need to stop parents from responding to this event - which may cause a double lift
+    // We also need to NOT call event.preventDefault() so as to maintain as much standard
+    // browser interactions as possible.
+    // event.preventDefault() in an onTouchStart blocks almost
+    // every other event including force press
+    event.stopPropagation();
 
     startPendingDrag(event);
   };
--- a/src/view/draggable/connected-draggable.js
+++ b/src/view/draggable/connected-draggable.js
@@ -57,6 +57,12 @@ const defaultMapProps: MapProps = {
   direction: null,
 };
 
+// $ExpectError - using spread
+const preLiftMapProps: MapProps = {
+  ...defaultMapProps,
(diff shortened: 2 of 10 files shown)


src/view/draggable/connected-draggable.js
```diff
@@ -216,6 +222,11 @@ export const makeSelector = (): Selector => {
         };
       }
 
+      // a lift is in progress - do not let anything start a lift
```
alexreardon: This is unrelated but I thought of adding it @jaredcrowe as it is a bit safer. It does cause another render in the lift cycle which is lame - but I think it is safest to do this

src/view/drag-handle/sensor/create-touch-sensor.js
```diff
@@ -263,8 +263,12 @@ export default (callbacks: Callbacks, getDraggableRef: () => ?HTMLElement): Touc
...
     }
 
-    // Not stopping the event as we want to allow force press and other events to occur
-    // Scrolling is stopped by onTouchMove
+    // We need to stop parents from responding to this event - which may cause a double lift
+    // We also need to NOT call event.preventDefault() so as to maintain as much standard
+    // browser interactions as possible.
+    // event.preventDefault() in an onTouchStart blocks almost
+    // every other event including force press
+    event.stopPropagation();
```
jaredcrowe: this is the fix?

alexreardon: yep - it stops the event bubbling up.
it is super important that we do not call prevent default here - but stopPropogation is fine

Issue #166, opened 2017-11-07
Move to fuzzy dependency versions


leipert:
The package.json pins down dependencies pretty hard. This leads to duplicate dependencies in our project, e.g. `prop-types` which already has newer releases. Would you consider to updating all dependencies with carets `^` or at least tildes `~`? Or is there any reason why pinning them down, if you rely on yarn anyway.

PS: I see that you have travis-ci in place, you could think about adding [greenkeeper](https://greenkeeper.io/) as well to update dependencies automatically and report breakages with newer versions of this package.

alexreardon:
Hi @leipert.
I will ask around if there are any potential implications of moving to fuzzy dependencies. Keeping the dependencies locked adds a large level of confidence as we can test all the code exactly as it will be given to users. Even if we lock the dependencies in the lock file it seems like you would like to use dependencies that are not the exact same version as would be in the lock file - rather moving to a fuzzy range. Adding a fuzzy range potentially introduces bugs if dependencies break in minor or patch versions. I would hope that the dependencies of the library are light enough that this would not really be an issue anyway. 

Regarding `prop-types`. I am actually thinking of removing them altogether and just leaning on flow to reduce overhead. Also, this dependency should move to ~0kb in production if you set up your build correctly. That is my understanding anyway. Please let me know if that is not the case

Regarding greenkeeper - it is awesome and I want to use it. Sadly we have had issues with latest dependencies not working together well. That was about two months ago so I might try to push to latest of everything soon. If we can get to latest then yes a tool like greenkeeper would be useful #37

alexreardon:
@treshugart do you see any issues with moving to fuzzy deps?

treshugart:
@alexreardon There could be unexpected breakages for consumers if a dep is broken. I think it's okay to be optimistic here, though, especially in open source land, so +1 for fuzy.

leipert:
👍  Nice work, thank you.

alexreardon:
We also added greenkeeper @leipert to help going forward. This was not added previously due to an eslint issue (see #37 for details)

PR #178, opened 2017-11-16
⬆️ Upgrade raf-schd to 2.0.2

IanVS:
I have recently started to use react-beautiful-dnd, and found that I was getting flow errors from raf-schd:

```
Error: node_modules/raf-schd/test/index.spec.js:2
  2: import { replaceRaf } from 'raf-stub';
                                ^^^^^^^^^^ raf-stub. Required module not found
```

It seems that version 2.0.1 was publishing its test files to npm.  This version fixes that problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "memoize-one": "3.0.1",
     "prop-types": "15.5.10",
     "prop-types-extra": "1.0.1",
-    "raf-schd": "2.0.1",
+    "raf-schd": "2.0.2",
     "react-motion": "0.5.0",
     "react-redux": "5.0.6",
     "redux": "3.7.2",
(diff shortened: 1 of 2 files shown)


IanVS:
Out of curiosity, why are you using fixed versions _and_ `yarn.lock`?  It's a bit unfortunate, since if you had been using a caret, this PR wouldn't have been necessary and I wouldn't have gotten flow errors in my project when I tried to use react-beautiful-dnd.  I'll make the change as requested, I'm just curious what your motivation is.

alexreardon:
It was to increase stablity. Ideally the internal dependencies are small and private. However, it looks like we will probably move to fuzzy versions #166. For now I asked if you could keep it as the exact version for consistency until we make the jump

Cheers

alexreardon:
Thanks for this @IanVS - it will go out in the next release

Issue #37, opened 2017-08-16
Upgrade to the latest version of flowtype


alexreardon:
Details:
https://medium.com/flow-type/even-better-support-for-react-in-flow-25b0a3485627

/cc @thejameskyle how would this change play with your prop type generator?

thejameskyle:
I need to rework it to lookup `Props` from the type parameter instead of from the class property.

alexreardon:
I am thinking of dropping prop types... not sure

alexreardon:
https://flow.org/en/docs/react/

alexreardon:
I tried to upgrade, but sadly I am unable to until eslint is fixed: https://github.com/yannickcr/eslint-plugin-react/issues/1376

thejameskyle:
https://github.com/yannickcr/eslint-plugin-react/issues/1376#issuecomment-325101762

alexreardon:
I think we should be good to start on this one now

alexreardon:
I tried to upgrade today but the eslint plugin is still yet to be fixed. No harm in waiting for now

alexreardon:
Closed by #181

PR #181, opened 2017-11-19
Upgrading all the deps

### LINKED ISSUE #166: Move to fuzzy dependency versions

alexreardon:
As the name says - this PR upgrades all the dependencies to the latest version (except for react).

Previously this was blocked by #37 

- [x] All deps
- [x] eslint to latest
- [x] flow to latest
- [x] Move from npm ignore file to files array in package json (moving from black listing to white listing)
- [x] Removing (unused) prop type generation
- [x] Moving to fuzzy dependencies (closes #166)

I also updated the `ISSUE_TEMPLATE` as per our discussion @jaredcrowe


.npmignore
```diff
@@ -1,37 +0,0 @@
-# Editors
```
alexreardon: Now moved to whitelisting in package.json

src/view/droppable-dimension-publisher/droppable-dimension-publisher.jsx
```diff
@@ -47,10 +45,12 @@ export default class DroppableDimensionPublisher extends Component {
       isDropDisabled,
       targetRef,
     } = this.props;
-    invariant(targetRef, 'DimensionPublisher cannot calculate a dimension when not attached to the DOM');
```
jaredcrowe: why get rid of invariant here but not elsewhere in this file where it's used?

alexreardon: Good call - i will look at why this was done. I think i was battling a flow issue.

package.json
```diff
@@ -15,6 +15,9 @@
     "url": "https://github.com/atlassian/react-beautiful-dnd/issues"
   },
   "main": "lib/index.js",
+  "files": [
+    "/lib"
```
jaredcrowe: publishing this now?

src/view/droppable-dimension-publisher/droppable-dimension-publisher.jsx
alexreardon: I think it has to do with flow not understanding invariant

package.json
alexreardon: yeah - just the lib 💃 
I did a test run and it looks good

jaredcrowe: oh this is because we got rid of the .npmignore?

src/view/draggable/draggable-types.js
```diff
@@ -109,18 +109,18 @@ export type StateSnapshot = {|
 // Using PropType<*,*> to allow strict typing within code and looser typing
```
jaredcrowe: can probably remove this comment now?

package.json
alexreardon: you can run `yarn pack` to see what will be published

src/view/draggable/draggable-types.js
alexreardon: nice. done

package.json
```diff
@@ -29,50 +32,45 @@
     "prepublish": "yarn run build"
   },
   "dependencies": {
-    "babel-plugin-react-flow-props-to-prop-types": "0.15.0",
-    "babel-runtime": "6.26.0",
-    "invariant": "2.2.2",
-    "memoize-one": "3.0.1",
-    "prop-types": "15.5.10",
```
jaredcrowe: I don't think we can remove prop-types, it's still used in some places... (typing context types for instance)

package.json
```diff
@@ -29,50 +32,45 @@
...
-    "jest": "21.1.0",
-    "raf-stub": "2.0.1",
-    "react": "15.6.1",
-    "react-dom": "15.6.1",
-    "react-test-renderer": "15.6.1",
-    "rimraf": "2.6.2",
-    "styled-components": "2.1.2"
-  },
-  "peerDependencies": {
-    "react": ">=15.4.0"
```
jaredcrowe: keep react as a peerdep rather than a devdep?

alexreardon: Kept it as both - i think we needed it for storybook

package.json
alexreardon: Correct. I have re-added it

package.json
```diff
@@ -29,50 +32,46 @@
     "prepublish": "yarn run build"
   },
   "dependencies": {
-    "babel-plugin-react-flow-props-to-prop-types": "0.15.0",
```
jaredcrowe: we don't need this any more?

alexreardon: This is probably the most contentious part of the PR. This is used to generate prop types from flow. However, it is just one more thing to maintain and makes it a little hard to move forward quickly. My thought is to drop it given that we publish flow types and there is public typescript types also. 

What are your thoughts?

alexreardon: Update: we were not generating prop types anyway. Given that nobody has complained I think we just leave it as is

Issue #142, opened 2017-10-19
Distribution files?


Frozenlock:
In the Clojurescript world we often [package JS libraries](https://cljsjs.github.io/) for ease of use.

When doing so, [it's encouraged](https://github.com/cljsjs/packages/wiki/Creating-Packages#2-find-the-librarys-source-code) to use the official distribution files (`dist/`). This helps us upgrade packages and keep them in sync with the JS libraries.

I was ready to package and use this project but couldn't find those files :cry: 

Would it be possible to add this into react-beautiful-dnd?

jaredcrowe:
Hey @Frozenlock thanks for raising this. We investigated this in the past (#81) and it's something we'd be happy to add at some point. Right now we are focused on priorities from our immediate roadmap so we won't be able to address this right away, but we'd be open to PRs from anyone from the community who would like to work on this :)

vinifala:
I'd like to contribute to solve this issue. However, I am not entirely sure how to approach it. At first I thought to remove `./lib` from git ignore. But after brief research, looks like Clojurescript expect a bundle. Would a rollup/webpack bundle be a viable option?

alexreardon:
Great to hear @vinifala!! I am not familiar with the particulars of what is needed for Clojurescript. The ideal would be if we also built a single bundle for distribution. Eg in `/dist`

alexreardon:
Please take a look at the prior work here: #81

alexreardon:
Closed by #186

alexreardon:
@Frozenlock let me know how you go getting it onto https://cljsjs.github.io/ . I can add a link from this repo when it is up. Cheers

alexreardon:
@Frozenlock any update on this?

Frozenlock:
@alexreardon Sorry for the delay, I'll try to do it in the next few days.

alexreardon:
All good!

Frozenlock:
Done!
React-beautiful-dnd is now available as a [cljsjs package](https://cljsjs.github.io/).

Thank you so much for the distribution files!

alexreardon:
So great @Frozenlock !!! I'll add a link to the readme!

PR #186, opened 2017-11-20
chore(build): add distribution files

vinifala:
Added distribution bundle.
https://github.com/atlassian/react-beautiful-dnd/issues/142

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,28 +3,21 @@
   "version": "2.6.3",
   "description": "Beautiful, accessible drag and drop for lists with React.js",
   "author": "Alex Reardon <areardon@atlassian.com>",
-  "keywords": [
-    "drag and drop",
-    "dnd",
-    "react",
-    "react.js",
-    "natural",
-    "beautiful"
-  ],
+  "keywords": ["drag and drop", "dnd", "react", "react.js", "natural", "beautiful"],
   "bugs": {
     "url": "https://github.com/atlassian/react-beautiful-dnd/issues"
   },
   "main": "lib/index.js",
-  "files": [
-    "/lib"
-  ],
+  "files": ["/lib", "/dist"],
   "scripts": {
     "test": "jest --config ./jest.config.js",
     "validate": "yarn run lint && yarn run typecheck",
     "lint": "eslint .",
     "typecheck": "flow check",
-    "build": "yarn run build:clean && yarn run build:lib && yarn run build:flow",
+    "build":
+      "yarn run build:clean && yarn run build:lib && yarn run build:flow && yarn run build:dist",
     "build:clean": "rimraf lib",
+    "build:dist": "rollup -c && rollup -c --min",
     "build:lib": "babel src -d lib",
     "build:flow": "flow-copy-source --verbose src lib",
     "storybook": "start-storybook -p 9002",
@@ -71,6 +64,11 @@
     "react-dom": "^15.6.1",
     "react-test-renderer": "^15.6.1",
     "rimraf": "^2.6.2",
+    "rollup": "^0.51.8",
+    "rollup-plugin-babel": "^3.0.2",
+    "rollup-plugin-commonjs": "^8.2.6",
+    "rollup-plugin-node-resolve": "^3.0.0",
+    "rollup-plugin-uglify": "^2.0.1",
     "styled-components": "^2.2.3"
   },
   "license": "Apache-2.0",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,0 +1,23 @@
+import resolve from 'rollup-plugin-node-resolve';
+import commonjs from 'rollup-plugin-commonjs';
+import uglify from 'rollup-plugin-uglify';
+import babel from 'rollup-plugin-babel';
+import yargs from 'yargs';
+
+const args = yargs.argv;
+const external = ['react'];
+
+export default {
+  input: './lib/index.js',
+  output: {
+    file: `dist/index${args.min ? '.min' : ''}.js`,
+    format: 'umd',
+  },
+  name: 'ReactBeautifulDnd',
+  plugins: [babel({ exclude: 'node_modules/**', babelrc: false }), resolve(), commonjs()].concat(
+    args.min ? uglify() : []
+  ),
+  sourceMap: false,
+  external,
+  globals: { react: 'React' },
+};
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,3 +1,4 @@
 # Generated files
+dist/
 lib/
 flow-typed/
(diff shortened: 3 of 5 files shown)


alexreardon:
Does this naming convention satisfy the Clojurescript requirements? @Frozenlock?

Frozenlock:
Thank you for doing this.

IIRC as long as one can easily generate a minified file for distribution we can package it with a minimum of manual operations.

I'll try to look at this in more details this evening and come back with an answer.

alexreardon:
I think we also need to ensure that we are bundling react

alexreardon:
Also, there is a lot of webpack stuff in there that we could avoid if we used [rollup](https://rollupjs.org/)

Frozenlock:
After reviewing the process, the thing we need is the browserified file (like this [one](https://unpkg.com/react-datepicker@0.53.0/dist/)).

We provide our own packaged dependencies for the most popular libraries (React, react-motion, etc.).
This means that if they are in the _externals_ section of the webpack.config.js, we can avoid duplication.

However, I don't know enough about the vanilla JS ecosystem nor do I want to burden this project.
I would simply ask that you do distribution files as you normally would; we'll handle it from there.

vinifala:
- [x] Swap out Webpack for Rollout
- [x] add dist to package.json#files
- [x] add dist to .gitignore
- [x] remove the dist directory from this commit

This is my first time using Rollup, I followed the implementation mentioned in https://github.com/atlassian/react-beautiful-dnd/pull/81

alexreardon:
This is looking good. @treshugart does it look good to you?

@Frozenlock would this satisfy your requirements?

vinifala:
Thank you @alexreardon. Please let me know if that suits your bill, @treshugart @Frozenlock :)

package.json
```diff
@@ -15,16 +15,16 @@
     "url": "https://github.com/atlassian/react-beautiful-dnd/issues"
   },
   "main": "lib/index.js",
-  "files": [
-    "/lib"
-  ],
+  "files": ["/lib", "/dist"],
```
treshugart: Would be good to also add a `browser` field that maps to `dist/index.js` (it will be minified by bundlers, however unpkg won't serve the min file).

treshugart:
LGTM but left a comment that we might think about adding now. Some things I'd suggest thinking about for the future:

- `dist/` is really `umd/` and can map to `browser` (I've commented about mapping the current `dist/` to that)
- `lib/` is really `node/` and can map to `main`
- `es/` that exports es5 (or lowest common denominator) + esm and can map to `module`
- Having an `esnext/` build for consuming the latest spec is something people are doing more and more. I've done this in Skate and map it to the `esnext` field.

alexreardon:
@treshugart can we implement your suggestions in a follow up issue? It seems like the currently PR solves the immediate need

treshugart:
> Some things I'd suggest thinking about for the future:

treshugart:
Or if you mean the `browser` field, that's up to if you think people might be reaching into the dist manually and if you want to consider a folder rename a breaking change. I'm don't care what you do here.

rollup.config.js
```diff
@@ -0,0 +1,27 @@
...
+export default {
+  input: './lib/index.js',
+  output: {
+    file: `dist/index${args.min ? '.min' : ''}.js`,
+    format: 'umd',
+  },
+  name: 'ReactBeautifulDnd',
+  plugins: [
+    resolve({
+      browser: true,
```
alexreardon: why is this needed?

```js
// some package.json files have a `browser` field which
      // specifies alternative files to load for people bundling
      // for the browser. If that's you, use this option, otherwise
      // pkg.browser will be ignored
      browser: true,  // Default: false
```

alexreardon: it looks like this already defaults to true

```js
// use "main" field or index.js, even if it's not an ES6 module
      // (needs to be converted from CommonJS to ES6
      // – see https://github.com/rollup/rollup-plugin-commonjs
      main: true,  // Default: true
```

rollup.config.js
```diff
@@ -0,0 +1,27 @@
...
+      main: true,
+    }),
+    commonjs({
```
alexreardon: not sure what this one is for. I think just listing react as an external is enough

vinifala: `rollup-plugin-commonjs` complains that React does not export `'Component', 'Children', 'createElement'` as named exports. This was the solution I've found.

vinifala: Listing React as an external dependency leads to
`(!) Missing global variable name
Use options.globals to specify browser global variable names corresponding to external modules
react (guessing '_react')`
I'm not sure if that's fine, or should I include `globals: { react: 'React' }` to `rollup.config.js`?

treshugart: Yeah, I think you're going to have to do that. I've found you can automate the dependencies part by doing something like:

```js
const pkg = require('./package.json');
const skip = {
  ...pkg.dependencies,
  ...pkg.devDependencies,
  ...pkg.optionalDependencies,
  ...pkg.peerDependencies
};
```

Jason Miller (of Preact fame) also shows this [here](https://gist.github.com/developit/41f088b6294e2591f53b).

alexreardon:
This is looking good! I will give it a look over next week.

vinifala:
Thanks @alexreardon

rollup.config.js
```diff
@@ -0,0 +1,28 @@
+import resolve from 'rollup-plugin-node-resolve';
+import commonjs from 'rollup-plugin-commonjs';
+import uglify from 'rollup-plugin-uglify';
+import babel from 'rollup-plugin-babel';
+import yargs from 'yargs';
+
+import pkg from './package.json';
+
+const args = yargs.argv;
+const external = Object.keys(pkg.dependencies || {}).concat(
```
alexreardon: Getting close! I think this should only include `react`. The other dependencies are internal and should not need to be provided by consumers.

Also, I am getting this warning:

```
(!) Missing global variable names
Use options.globals to specify browser global variable names corresponding to external modules
react (guessing '_react')
```

I would expect that this should be listed as 'React' in the `options.globals`

alexreardon:
Thanks for preserving with this @vinifala

vinifala:
Thanks @alexreardon, React added as only external dependency with proper global mapping :)

vinifala:
`oci runtime error: exec failed: container_linux.go:265: starting container process caused "could not create session key: disk quota exceeded"`
Looks like we ran out of space.

alexreardon:
Oh wow! I am not too worried about that at this stage

alexreardon:
Well done @vinifala !! I had a look locally and it is looking good. I'll merge this pr and it will go out in the next release. Thanks for your ongoing efforts to get this across the line 🎉

vinifala:
Thanks for your patience, @alexreardon. I'm looking forward to contribute further.

PR #195, opened 2017-11-23
Improved default for interactive element

### LINKED ISSUE #190: Improved default for child interactive elements
Currently [we suggest monkey patching](https://github.com/atlassian/react-beautiful-dnd#interactive-child-elements-within-a-draggable) event handlers to block dragging on interactive elements. This was a simple technique based on the proposition that not many people would want interactive elements within their drag handles anyway. However, this has been a constant source of issues for the repo.

#189, #185, #149, #110 (there are probably others)

I propose that we block dragging on child interactive elements by default - with an ability to opt out. If a consumer does opt out they are still welcome to introduce their own monkey patching logic for more complex scenarios. I think this would help the library 'just work' for more people out of the box.

If the type of the `event.target` (the element actually interacted with - `event.currentTarget` is the one with the listener [more](https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget)) is one of the following nodes, then a drag is prevented by default:

- `input`
- `textarea`
- `select`
- `option`
- `button`

The consumer could opt out this by adding a prop to `Draggable`. Something like `disableInteractiveChildrenBlocking` (name pending 👍 ). They could then do monkey patching if they liked

alexreardon:
Closes #190 

It feels much better and less broken by default for these elements. Currently based off the moving-to-long-press branch to avoid conflicts. This also cleans up the docs a lot more as there is less you need to know.

I am thinking of releasing this as a `feature` change 0.x as I have come to see that the current behaviour can be considered a bug - it rendered interactive elements unusable by default. Adding it will not break those who have monkey patched the event handlers - it is just redundant


src/view/drag-handle/util/should-allow-dragging-from-target.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+import type { Props } from '../drag-handle-types';
+
+export const interactiveTagNames: string[] = [
+  'input',
+  'textarea',
+  'select',
+  'option',
+  'button',
+  'contenteditable',
```
bradleyayers: `contenteditable` isn't a tag name, it's an attribute that can be applied to elements.

alexreardon:
Cheers

alexreardon:
@bradleyayers I have not played with `contenteditable` before 😄 

It looks like it is an attribute that can be applied to an element that makes all of the content within it editable. This element can itself contain other elements. This makes things a little more complicated because we no longer need to just check the `event.target` itself, but we would also need to walk up the tree to the `event.currentTarget` to see if any of the parents had a `contenteditable` attribute. Does my understanding sound correct to you?

alexreardon:
@bradleyayers this seems to be working how I expect. I will update the tests. 

```js
// @flow
import type { Props } from '../drag-handle-types';

export const interactiveTagNames: string[] = [
  'input',
  'textarea',
  'select',
  'option',
  'button',
];

const isContentEditable = (parent: HTMLElement, current: ?HTMLElement): boolean => {
  if (current == null) {
    return false;
  }

  if (current.getAttribute('contenteditable') === 'true') {
    return true;
  }

  // nothing more can be done and no results found
  if (current === parent) {
    return false;
  }

  return isContentEditable(parent, current.parentElement);
};

export default (event: Event, props: Props): boolean => {
  // Allowing drag with all element types
  if (props.canDragInteractiveElements) {
    return true;
  }

  const isTargetInteractive: boolean =
    interactiveTagNames.indexOf(event.target.tagName.toLowerCase()) !== -1;

  if (isTargetInteractive) {
    return false;
  }

  // Need to see if any element between event.target and event.currentTarget (inclusive)
  // is contenteditable. A contenteditable element can contain other elements
  return !isContentEditable(event.currentTarget, event.target);
};
```

bradleyayers:
I’m not sure exactly what the event target will be, I’d have to do some experiments.

alexreardon:
The target is the element you interact with - even if it is the child

https://codepen.io/alexreardon/pen/OOERGN?editors=1011#0

src/view/drag-handle/util/should-allow-dragging-from-target.js
```diff
@@ -0,0 +1,26 @@
+// @flow
+import type { Props } from '../drag-handle-types';
+
+export const interactiveTagNames: string[] = [
```
jaredcrowe: Add `audio`, `video` and `optgroup` as per sean's suggestion

alexreardon: done :)

alexreardon:
@jaredcrowe I just pushed up a story :)

jaredcrowe:
looks like we're having some flow issues but once those are resolved this looks good to go

alexreardon:
I am looking at them now

alexreardon:
Resolved @jaredcrowe 👍

PR #218, opened 2017-12-06
doc: minor typo: where update to were

ajaymathur:
updated 

*more than they would if they **where** just* 

to 

*more than they would if they **were** just*

---

❤️ the docs. Easily the best documented library I have come across.


alexreardon:
Thanks @ajaymathur !

Issue #223, opened 2017-12-11
Conditional logging


alexreardon:
Currently we do a lot of logging to the console in the case of errors. However, this is a lot of strings to ship to production. My idea is to create a `log` module which will only log to the console if `process.env.NODE_ENV !== 'production'`. That way these logs can be tree shaken out for production.

The draw back would be no error logging in production which would make problems harder to diagnose.

ajaymathur:
If ok, I can work on this over the week end. 🙂

ajaymathur:
Idea ( not sure if it would be the best way ) :  how about creating a babel plugin to remove `console.log` when NODE_ENV is production (not case sensitive) and in case if we really need any log to be in production we can update syntax as `console.log('#PROD: <message>')` , plugin will only remove #PROD: from this message, this way we will only remove code for prod and not add any line of code as all.

Thoughts 💭 ?

ajaymathur:
This will give us some control over console logs, in case we want to replace them with api calls to log failures, 🤔

alexreardon:
I think that will be too heavy to maintain for this project. My ideal is just to have a module that exposes a 'warn' and 'error' function that only logs to console if `process.env.NODE_ENV !== 'production'`. There are existing tree shaking tools to remove this code when then this conditional is not met.

Something like:

```js
export const warn = (...args: mixed[]) {
  if(process.env.NODE_ENV !== 'production') {
    console.warn(...args);
  }
}
```

alexreardon:
The main purpose of this is to reduce the production bundle size

If it is too difficult then we won't worry about it. We should only be logging in error scenarios

alexreardon:
Done as part of #255

PR #225, opened 2017-12-14
added new logger module and update all console logging to be done thr…

ajaymathur:
…ough that module

1. added a logger module ( TODO: Better name for module directory maybe utils? )
2. removed no-console rule from eslintrc
3. updated all test cases accordingly

😃 

Related to https://github.com/atlassian/react-beautiful-dnd/issues/223


.eslintrc
```diff
@@ -82,9 +82,6 @@
     // Not allowed: `const _secret = Symbol()`
     "no-underscore-dangle": ["error", { "allowAfterThis": true }],
 
-    // Allowing warning and error console logging
-    "no-console": ["error", { "allow": ["warn", "error"] }],
```
alexreardon: Let's keep this rule and just add an opt out for the log file

src/logger/logger.js
```diff
@@ -0,0 +1,33 @@
...
+  }
+}
+
+export function error(...args: mixed[]) {
+  if (process.env.NODE_ENV !== 'production') {
+    console.error(...args);
+  }
+}
+
+export function group() {
```
alexreardon: group should take a title (or at least an optional one)

src/state/action-creators.js
```diff
@@ -18,6 +18,7 @@ import type {
 import noImpact from './no-impact';
 import getNewHomeClientCenter from './get-new-home-client-center';
 import { add, subtract, isEqual } from './position';
+import { error } from '../logger/logger';
```
alexreardon: For all the imports I think it will read better if we do:

```js
import * as log from '../logger/logger';

log.error('my error');
```

Otherwise it might look like we are calling an error handler

alexreardon:
Thanks heaps for this @ajaymathur !!!! ❤️

ajaymathur:
Tested with production build, logger functions are tree shaken. ( Its a cool thing I learned. Thanks ) 😸 

image

alexreardon:
@ajaymathur the big PR has now been merged. You should be good to do this now

ajaymathur:
Hi @alexreardon , I have updated the src files but not the tests .. I know couple of ways that we can mock in tests, I need to know to which one do you think is better and possibly reasoning behind it (For my learning 🙂 )

- Mock logger modules. 
```js
import * as logger from '../log';
jest.mock('../log');

// in afterEach, in all describes needed

jest.resetAllMocks()
```

- Mock each function one by one,
```js
import * as logger from '../log';

// in before each
loggerError = jest.spyOn(logger, 'error').mockImplementation(() => { });

// in after each
 loggerError.mockRestore();

// downside function name is changed
```

- Replace functions as is, console with logger, might need flow fix (update types maybe) for this.
```js
import * as logger from '../log';

// in before each
 jest.spyOn(logger, 'error').mockImplementation(() => { });

// in after each
logger.error.mockRestore();
```

I like the first way as it will be simple and will reset all mocks 🙂

ajaymathur:
sorry for late update .. 👨‍💻

jest.config.js
```diff
@@ -2,4 +2,8 @@
 
 module.exports = {
   setupFiles: ['./test/setup.js'],
+  clearMocks: true,
```
ajaymathur: https://facebook.github.io/jest/docs/en/mock-function-api.html#mockfnmockclear , it think clearMock is the one required. I got few error of mocks calls not getting reset properly between calls.

alexreardon: Interesting! Let's chat on monday

alexreardon:
Looking good @ajaymathur! Can you please change the eslint rule so that console.warn, error etc are disabled - then we add an opt out in the logger. This should hopefully encourage people to use the logger and not the console directly

test/unit/log.spec.js
```diff
@@ -0,0 +1,126 @@
+// @flow
+import * as log from '../../src/log';
+
+describe.only('log', () => {
```
alexreardon: we will need to remove .only

.eslintrc
```diff
@@ -83,7 +83,7 @@
     "no-underscore-dangle": ["error", { "allowAfterThis": true }],
 
     // Allowing warning and error console logging
-    "no-console": ["error", { "allow": ["warn", "error"] }],
+    "no-console": "error",
```
alexreardon: win

jest.config.js
```diff
@@ -2,4 +2,7 @@
 
 module.exports = {
   setupFiles: ['./test/setup.js'],
+  testMatch: [
```
alexreardon: do we need this?

ajaymathur: by default jest will look at `[ '**/__tests__/**/*.js?(x)', '**/?(*.)(spec|test).js?(x)' ]` we can narrow it down. I am not certain if it is perf improvement for test but  I have seen this config in most projects.

Issue #188, opened 2017-11-21
Remove prop `type` from `Draggable`


alexreardon:
Given that a `Draggable` must have the same `type` as it's parent `Droppable` - there is no point it requiring a `Draggable` to declare its `type`. We would need the prop if we allowed floating draggables #29. I suppose the answer to this question is whether we want to support floating draggables

alexreardon:
Related: #25 

Right now I am thinking of removing `type` from `Draggable` and using the `type` from the parent `Droppable`. This would remove the need for #25 and would also make #29 redundant.

If you really did need a 'floating' `Draggable` you could create a small `Droppable` that has `isDropDisabled` set to `true` to avoid any dropping on it. Given we are optimising for lists I think this is the best course of action.

Thoughts @jaredcrowe ?

alexreardon:
I had a crack at this - but I fell down in the `selector` function. We use the type in the selector to avoid needing to do any work on draggables that are not impacted by a drag. Selectors cannot read from context. More information: https://github.com/reactjs/react-redux/issues/289

Given this limitation I think the simplest path is to stick with the `type` prop for now. However, we can do #25 to make the developer experience nicer.

I am thinking of a way around this - but it feels gross. The idea would be we wrap what is the `Draggable` today which reads the type from the context and then passes that on as a prop. This would not impact the consumption story at all but introduces a level of indirection that could get confusing

alexreardon:
The `type` prop may no longer be needed as apart of #86. Hopefully as a part of that work we can drop the `type` from `Draggable` altogether 👍

alexreardon:
I can confirm that the `type` prop is being removed from `Draggables` as a part of #86

PR #226, opened 2017-12-14
Drastic improvements to performance

### LINKED ISSUE #86: Improve the performance of dimension capturing
Currently there is a `DIMENSION_COLLECTING` phase **before** a drag where we request the dimensions for all of the `Droppable` and `Draggable` components. This is quite expensive as it calls `.getBoundingClientRect` and `window.getComputedStyle` on all the nodes. We need to think more about how we could improve the performance of this. Some options off the top of my head:

- (simplier) find a cheaper way of getting the margin values off an element. I had a look and I could not find one, but it is worth further investigation
- (simplier) allow drag movement before all the dimensions are requested (this would not help with keyboard dragging)
- (more complex) only request a batch of dimensions (such as those in the current viewport) and request more dimensions as needed. This is tricky because it may jank the scroll, and also will change how we calculate the index of the current item.

alexreardon:
Closes #86 #188 

This will push the repo to 4.0. There is still a lot of work to do, but I thought it best to start track it here

- [x] tests for new dimension marshal
- [x] tests for new style marshal
- [x] tests for new draggable selectors
- [x] Investigate drop / lift bottleneck i have observed
- [x] tests for `canLift` context fn
- [x] tests for `canLift` blocking in drag handle
- [x] update drag-handle tests for new logic
- [x] remove old connected draggable and droppable publisher tests
- [x] swap order of functions that take (current, previous) to be (previous, current) to follow english conventions
- [x] add test to drag-handle to ensure it is applying the style context
- [x] fix quick reordering incorrect visuals
- [x] write tests for flushing and dimension capturing
- [x] improve global lifecycle styles
- [x] update all drag impact functions to use new visual filter
- [x] now that we have viewport filtering - disallow keyboard movements offscreen (temp until we support auto scrolling)
- [x] combine ClientRect and DimensionFragment into one type (Area)
- [x] remove redundant conversion functions between client rects and dimension fragments
- [x] fix mouse displacement when over foreign lists 
- [x] refactor in droppable functions to sit along side visibility checks
- [x] tests for droppable visibility filtering
- [x] tests for viewport filtering
- [x] tests for get-displacement
- [x] fix issue where holding over the bottom of a full list with a scroll container causes jumping. This is caused by the droppable thinking it is no longer being dragged over
- [x] move from is-partially-visible to is-visible-through-frame functions
- [x] tests for updated get-droppable-over placeholder buffer (this will also test the visibility flicker issue
- [x] add test to get-droppable-over to ensure that it ignores disabled droppables
- [x] tests for new connected draggable
- [x] tests for new connected droppable
- [x] reordering big columns seems to not be working. ignoreContainerClipping? or is the center point not 'visible' and so being excluded?
- [x] ensure run time disabling of droppables still works correctly
- [x] tests for droppable placeholder
- [x] tests for new draggable publisher
- [x] tests for new droppable publisher
- [x] tests for refactored drag-impact functions
- [x] tests for get best cross axis refactor
- [x] droppable-dimension-publisher publishing isEnabled changes
- [x] audit all dimension choices (withMargin etc) to ensure that they are the same as previous. Also ensure this is tested
- [x] tests for animation configuration in connected-draggable
- [x] update exisiting tests
- [x] update examples to use new props
- [x] write docs explaining new props
- [x] write docs explaining recommended consumer performance optimisation for `DroppableStateSnapshot > isDraggingOver`
- [x] write release notes
- [x] audit all TODO comments
- [x] ensure there are no linting or flow issues
- [x] cross browser test
- [x] mobile test
- [x] fix regression found by @jaredcrowe 

@jaredcrowe I will let you know when it is ready for review - but it will not be until next week at the earliest


jaredcrowe:
Noticed a strange regression. Steps to replicate:

- View the [vertical list stress test story](https://deploy-preview-226--react-beautiful-dnd.netlify.com/?selectedKind=multiple%20vertical%20lists&selectedStory=stress%20test&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel).
- Drag an item into a list from another list which appears 'after' it. i.e. dragging into Delta from Kappa, or into Beta from Gamma will trigger this, but not into Beta from Alpha.
- Try to lift this item again - it will be impossible to lift with either the mouse or keyboard.
- Drag another item into the list - it will animate as if the previously moved draggable isn't in the list. (This is most noticeable when the previous item was dropped into the last position in the list)

alexreardon:
@jaredcrowe issue has been resolved and tests added

jaredcrowe:
Sorry dude, but I've noticed a couple of regressions with keyboard dragging 😿 These are both noticeable when playing with the [multiple vertical stress test story](https://deploy-preview-226--react-beautiful-dnd.netlify.com/?selectedKind=multiple%20vertical%20lists&selectedStory=stress%20test&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel).

**Issue 1:**
* Lift an item from Zeta (or drag an item from any other list into Zeta) using the keyboard.
* Try to move left into Delta or Epsilon. It will be impossible.
* This also occurs when trying to move from Delta or Epsilon to Alpha or Beta (but strangely, not to Gamma...)

**Issue 2:**
* Lift an item in Iota and move it down beyond the extend of Kappa.
* Move right into Kappa. The item will move _below_ the last visible draggable in Kappa rather than _above_ it.

src/state/move-to-next-index/in-foreign-list.js
```diff
@@ -1,32 +1,37 @@
...
-import { isPointWithinDroppable } from '../is-within-visible-bounds-of-droppable';
 import { patch } from '../position';
 import moveToEdge from '../move-to-edge';
+import getDisplacement from '../get-displacement';
+import getViewport from '../visibility/get-viewport';
+import isVisibleInNewLocation from './is-visible-in-new-location';
 import type { Edge } from '../move-to-edge';
 import type { Args, Result } from './move-to-next-index-types';
-import type {
+import type
```
jaredcrowe: weird newlines here?

alexreardon: Yeah, lame auto import fix. I'll fix this one - but it should probably be an eslint rule eventually

alexreardon:
Issue 1 is fixed and tests added (invisible subject checks added)

Issue 2 is a regression - however, the previously experience was not great anyway. I would like to keep it as is because even if i fix it will need to be changed again for auto scrolling.

src/state/move-to-next-index/in-foreign-list.js
alexreardon: fixed

stories/src/vertical/quote-app.jsx
```diff
@@ -77,12 +70,11 @@ export default class QuoteApp extends Component<Props, State> {
 
   componentDidMount() {
     // eslint-disable-next-line no-unused-expressions
-    injectGlobal`
-      body.${isDraggingClassName} {
-        cursor: grabbing;
-        user-select: none;
-      }
-    `;
+    // injectGlobal`
```
jaredcrowe: comment

alexreardon: removed

PR #241, opened 2018-01-02
Allow boolean fields to be draggable

agirton:
`MouseDown` event is fired when the user taps on the checkbox. This works fine until the user drags the row and then once they do they have to click on the checkbox twice before it will show as checked.

Instead of creating an issue, I figured I would open with one with a PR to fix. **Note:** I'm not sure this is the best route or even the most scalable. The part I'm not crazy about is that it adds further validation logic based to find out what type of specific element it is. This was my first pass and thought it would best to get feedback before digging deeper.

I haven't created any new tests yet until it's determined if this is the route the project would like to take.

### Expected behavior
After dragging the item and then checking the checkbox I expect to see the checkbox checked. 😄 

### Actual behavior
If drag an element and then check the checkbox, it fires an event that it was checked. However, the UI is not showing as checked. If I click on the checkbox again it then shows up properly.

### Workaround
Until we find a solution I can set `disableInteractiveElementBlocking` to `true`.

### Steps to reproduce
1. Toggle the checkbox to see it works properly
2. Drag any element
3. Try toggling the checkbox to either uncheck/check and see that the check is not showing.
4. Click on checkbox again and see that it was properly propagated.

### Browser version
Tried in Latest Safari, Firefox, Chrome

### Demo
https://codesandbox.io/s/8lk2wo0809

checkbox-issue

### DIFF
--- a/src/view/drag-handle/util/should-allow-dragging-from-target.js
+++ b/src/view/drag-handle/util/should-allow-dragging-from-target.js
@@ -35,6 +35,14 @@ const isContentEditable = (parent: Element, current: ?Element): boolean => {
   return isContentEditable(parent, current.parentElement);
 };
 
+const isBooleanTag = (current: ?Element): boolean => {
+  if (current == null) {
+    return false;
+  }
+
+  return current.type === 'checkbox' || current.type === 'radio';
+};
+
 export default (event: Event, props: Props): boolean => {
   // Allowing drag with all element types
   if (props.canDragInteractiveElements) {
@@ -51,7 +59,7 @@ export default (event: Event, props: Props): boolean => {
   const isTargetInteractive: boolean =
     interactiveTagNames.indexOf(target.tagName.toLowerCase()) !== -1;
 
-  if (isTargetInteractive) {
+  if (isTargetInteractive && !isBooleanTag(target)) {
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


alexreardon:
Thanks for raising this one! I think we will need to do a little more thinking about it. We also need to think about whether clicking on the label should allow the default to happen. Often a checkbox is wrapped in a label and the whole thing is clickable as the toggle itself.

agirton:
I agree. I also saw this behavior when the checkbox is wrapped in a label.

alexreardon:
Sorry for the silence on this. I am super keen to see it happen

alexreardon:
I am surprised this is not picked up in our standard check which looks for whether the tagName is input. My understanding is that radio and checkbox elements must be:

- `<input type="checkbox">` OR
- `<input type="radio">`

I'll take a look

alexreardon:
I have added some examples that use radio and checkbox inputs. It seems to be working as intended.

https://react-beautiful-dnd.netlify.com/?selectedKind=nested%20interative%20elements&selectedStory=stress%20test&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel

I think you issue was caused by not applying the draggable style correctly

https://codesandbox.io/s/qxz19yy9q

PR #245, opened 2018-01-05
Move react and react-dom into peerDependencies

Li0liQ:


package.json
```diff
@@ -82,6 +80,10 @@
     "rollup-plugin-uglify": "^2.0.1",
     "styled-components": "^2.2.3"
   },
+  "peerDependencies": {
```
alexreardon: These where peer dependencies at one point. I am not sure when they where removed :|
I remember because we had fuzzy versions #84 

Also, we do not need the react-dom peer-dependency. It is not used in the production bundle

alexreardon:
Thanks @Li0liQ !!!

package.json
Li0liQ: Done.

Li0liQ:
https://github.com/atlassian/react-beautiful-dnd/pull/181 - the PR that removed react as from peerDepenendencies.

alexreardon:
Great pick up!!!

PR #267, opened 2018-01-15
Fixes 'flowtype' URLs

mrowles:
Flow URLs were missing protocol, so weren't linking to new Flow domain properly.


alexreardon:
Thanks @mrowles !!

PR #271, opened 2018-01-16
Fixing drag start lock

alexreardon:
Thanks @Li0liQ for finding this!


src/view/drag-handle/drag-handle.jsx
```diff
@@ -165,7 +168,7 @@ export default class DragHandle extends Component<Props> {
     }
 
     // this will check if anything else in the system is dragging
-    if (!this.canLift()) {
+    if (!this.canLift(this.props.draggableId)) {
```
alexreardon: This is the fix. Previously this allowed drags to start while a drop was occurring on the same element (which is not supported).

PR #278, opened 2018-01-17
simplification of drag handle scheduling

alexreardon:
A little cleanup. Also added some tests

### DIFF
--- a/src/view/drag-handle/util/create-scheduler.js
+++ b/src/view/drag-handle/util/create-scheduler.js
@@ -1,44 +1,32 @@
 // @flow
 import memoizeOne from 'memoize-one';
-import rafSchedule from 'raf-schd';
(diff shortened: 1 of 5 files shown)


src/view/drag-handle/util/create-scheduler.js
```diff
@@ -1,44 +1,32 @@
...
-  };
-
+export default (callbacks: Callbacks) => {
   const memoizedMove = memoizeOne((x: number, y: number) => {
     const point: Position = { x, y };
     callbacks.onMove(point);
   });
 
-  const move = rafSchedule((point: Position) => {
-    ifDragging(() => memoizedMove(point.x, point.y));
```
alexreardon: Rather than executing these checks every time, no we just cancel pending frames when they are no longer needed

Issue #211, opened 2017-11-30
Issue with bottom anchored lists


amytych:
When multiple draggable lists are anchored to the bottom of a container (e.g. using `display: flex` or `display: table`) there is an issue when dragging items between them. I suspect it is because of position calculations being based on the top left coordinate of the list. When the item is dropped, the placeholder element is removed from the DOM causing the top left coordinate to change. This results in the dropped item being briefly placed incorrectly, lower than it should, only then to revert to the proper position, as illustrated in the gif below:

![nov-30-2017 10-03-44](https://user-images.githubusercontent.com/998924/33422414-ee9be0ac-d5b5-11e7-8d7e-bf3f2fd45dab.gif)

Just to reiterate, everything works beautifully with regular lists, problem appears when they're attached to the bottom, that's the only change.

I tried many ways to work around this but didn't succeed. Was that scenario ever considered? Maybe someone was able to make it work? Any pointers would be greatly appreciated, even one saying that bottom anchored lists are out of scope of this library.

In any case, thank you!

alexreardon:
Bottom arranged lists had not crossed my mind! For now they are out of scope as you put forward. You could get something working by collapsing the droppable so that it tightly surrounds the list. This is not optimal as it would prevent the user from dropping on the visual space above. It might be super strange, but you could add an invisible Draggable to the top of each list that fills the space that has the `isDragDisabled` prop on it. It might be quite strange when it comes to movement though - but you could try it out!

edzis:
I am also in need for bottom-based lists, adding items to them and dragging between them. Will try out the idea of an empty placeholder.

edzis:
@alexreardon I tried several approaches with spacer element, `align-items: flex-end;`, placeholder order and for none of those I could get a combination of accurate placeholder logic and drag-end animation target.

Then explored the logic and modified the generated js in my test project. By introducing logic in a couple of places I managed to get further. Here is my result - https://imgur.com/a/dIuDB

This exploration is currently brutal but the changes seemed to be in places where configuration can easily alter it.

For my next exploration I want to find a way to change which droppables should be moved - the ones before the insert point. Do you have any recommendations or concerns about this approach?

PR #280, opened 2018-01-18
Add example for `stick to bottom`

edzis:
There are some glitches to resolve around such usage but this simple change works fine in most interactions.

bottom-based-lists-drag

Drop position calculations:
- are fine while container is larger than existing and incoming content
- are broken while content occupies the full container
- are fine again while content is larger than visible container 

The last point could be the key to fix also the second case.

Keyboard dragging has the same fine/broken scenarios.

### DIFF
--- a/stories/3-board-story.js
+++ b/stories/3-board-story.js
@@ -18,4 +18,7 @@ storiesOf('board', module)
   ))
   .add('long lists in a short container', () => (
     <Board initial={data.medium} containerHeight="300px" />
+  ))
+  .add('stick to bottom', () => (
+    <Board initial={authorQuoteMap} stickToBottom />
   ));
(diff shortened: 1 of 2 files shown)


alexreardon:
Thanks @edzis!!! This is very cool! However, given that it is not a supported behaviour right now and is a little broken I won't merge this PR. I would not want people to think that the behaviour is 100% supported when it has some broken parts.

Thank you for taking the time to make this PR. Hopefully we can look at truely supporting this behaviour soon!

edzis:
@alexreardon I do understand that this is not ready for public. But I was hoping on some hints about the scenario "while content is larger than visible container" in hopes that it could help me to prepare a proper fix for my own needs and for the benefit of other users (eg. #211). Please pardon the lack of clarity about the goal of this PR in the description.
Do you think you can provide such insights?

PR #308, opened 2018-02-03
Revert "added new logger module (#225)"

alexreardon:
This reverts commit a19aeab8b17da35fc40f5a73c422f66399cddde7.


alexreardon:
Reverting as it the change is not resulting in the log strings being removed from the production bundle. I think it is because we are using the import * from pattern. This might be a limitation of rollup

PR #309, opened 2018-02-05
strip console from builds

ajaymathur:
- using https://github.com/rollup/rollup-plugin-strip plugin to remove console from builds

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,6 +79,7 @@
     "rollup-plugin-commonjs": "^8.2.6",
     "rollup-plugin-node-resolve": "^3.0.2",
     "rollup-plugin-replace": "^2.0.0",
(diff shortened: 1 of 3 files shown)


alexreardon:
Great outcome @ajaymathur !!

Just to confirm everything is working, can you please confirm the size of the existing bundle compared to the new one?

To get the minified value - run yarn run build
To get the gzip value - right click on the file in macosx and click 'compress'. This is roughly correct

ajaymathur:
## Stats

| Matrix        | With logs (before) | After striping logs  | Reduction |
| ------------- |-------------| -----|---|
| min      | 141066 | 135245 |  5821 |
| gzip      | 37111  | 35726  | 1385 |

*all numbers in bytes

alexreardon:
Love it

PR #321, opened 2018-02-12
Auto scrolling (and more!)

alexreardon:
This PR contains:

- New accessibility improvements. Closes #311 
- New auto scrolling behaviour. Closes #27 
- New `onDragUpdate` hook. Closes #298 Closes #311 
- More information available in `DraggableStateSnapshot` and `DroppableStateSnapshot`. Closes #298 
- Maintain scroll location on drop. Adding a `data-*` hook to `Droppables` for future global styling. Closes #290

PR #325, opened 2018-02-14
update to babel-preset-env

alirezavalizade:


alexreardon:
We support a [range of browers]. We would need to instruct babel-preset-env about what browsers we are using using a browserlist. We could not just use the default.

Given that we support ie11 using babel-preset-2015 is sufficient for now. However, I am happy to proceed with this if you are willing to add the correct browser list 👍

alirezavalizade:
@alexreardon Please check [this](https://github.com/atlassian/react-beautiful-dnd/pull/325/commits/92b333bd9bb23db49db6a211f589bfcc81a50165).

alexreardon:
Hi @alirezavalizade, I think it is better to be a little more explicit to match our browser matrix. Here is what we are using in other products:

```
ie >= 11
last 1 Edge version
last 1 Firefox version
last 1 Chrome version
last 1 Safari version
last 1 iOS version
last 1 Android version
last 1 ChromeAndroid version
```
If you can add that we should be good to go!

alirezavalizade:
@alexreardon Hi, Great.
Thanks browser matrix!

alexreardon:
I'll check this tomorrow and ensure it is working correctly

Then we should be gewd

alexreardon:
I have not forgotten about this. I am just focusing on the moment on shipping #321. I hope to merge this for 5.0

alexreardon:
Whoops - i think i messed this up by trying to add a commit

alirezavalizade:
Thanks.

Issue #323, opened 2018-02-13
Way to reset data-react-beautiful-dnd-draggable for server side rendering 


czechdave:
## Bug or feature request?
Bug

### Expected behavior
`data-react-beautiful-dnd-draggable` shouldn't increment based on how many times it was run on the server but start fresh for each request.

### Actual behavior
Each time the server render is hit `data-react-beautiful-dnd-draggable` is incremented by 1.
i.e. after 6 reloads console throws:
```
Warning: Prop `data-react-beautiful-dnd-draggable` did not match. Server: "6" Client: "0"
```

### Steps to reproduce
Server side render and reload couple times, each reload the `data-react-beautiful-dnd-draggable` gets incremented by 1.

### Browser version
Chrome

https://github.com/atlassian/react-beautiful-dnd/blob/master/src/view/style-marshal/style-marshal.js#L11

Possible solution to this would be to expose a function that resets the counter that one could call during a server side render like so:
```
export function resetCounter() {
    count = 0;
}
```

alexreardon:
How interesting!! Sounds very useful indeed!

alexreardon:
The trouble will be exposing this api in a nice way.. Any reason your server is requiring the file multiple times?

czechdave:
I am not requiring the file multiple times. This happens because my express server loads once and then serves multiple requests. Each request needs to be handled in its own context but instead each is calling the same `count` variable.

alexreardon:
I am following now. Let's think about how we can solve this!

alexreardon:
My current thinking: an additional prop to `DragDropContext` to set the style context manually. However, this feels like a leak of an internal... hmmm

alexreardon:
@Blasz would you be keen to take a look at this one?

alexreardon:
Options:

1. remove the namespace (might be a bit lame when we support #302)
2. Provide an api option to `DragDropContext` to explicitly control the style context. This is probably the most flexible option - but it feels lame to expose this internal concern..
3. ?

Blasz:
@alexreardon sure!

czechdave:
I took a stab at this. This [PR #337](Add `resetContext` method for server side rendering #337) does exactly what I need to fix server side rendering though from your comments it seems this is not what you want.

PR #337, opened 2018-02-20
Add `resetContext` method for server side rendering

czechdave:
This fixes issue [#323](https://github.com/atlassian/react-beautiful-dnd/issues/323)


alexreardon:
Thanks for this @czechdave. @Blasz will be evaluating this approach soon

yogaboll:
This solved the problem for me. Thanks, @czechdave!

Blasz:
Hi @czechdave, I think your approach is the best way to fix #323 currently. As this is quite a small change I figured it would be easier to create a separate PR #345 that implements it in a slightly different way and adds unit/integration tests. I hope this is okay with you.
I'll therefore be closing this in favour of #345.

PR #340, opened 2018-02-22
Minor readme updates

Blasz:
Hey Alex,

I was going through the README tonight and noticed a few typos so thought I'd raise a PR to fix them.

I ended up re-ordering/restructuring some headers as well under draggable since I was finding it a bit hard to read and thought the headers would be better suited to be more accurate if displayed in a table of content.

I had a couple of questions after going through the doc that I'm not sure whether needs addressing or not.

1. Would it be worth adding a table of contents? The README is quite large, which is good, but I think a table of contents would allow easier navigation.
2. With regards to the droppable scroll container restrictions, is the window (html/body) excluded from being defined as a scroll parent? I noticed this distinction is made in the auto scroll section but not in the generic scroll container section
3. There is mention to both visible and visual draggable sibings. Are they interchangeable terms or should we stick with one?
4. I found the term Draggable Props to be a bit overloaded as it was being used for both the props that can be passed to the `Draggable` component itself and also the props coming out of the provided child function argument that are then passed to the Draggable node/inner ref element.
5. I found the usage of `DraggableProps > style` a bit obscure when referring to the style prop of DraggableProps, what do you think about using `DraggableProps.style` instead?

Am happy to update the PR in response to any of the questions above.

Cheers.


alexreardon:
Yes to everything!

alexreardon:
How soon could you action these? I am hoping to do a release this morning

Blasz:
I can try as soon as I get in, 30 - 60 mins

alexreardon:
perfect

PR #345, opened 2018-02-25
Add `resetServerContext` function for SSR

### LINKED ISSUE #323: Way to reset data-react-beautiful-dnd-draggable for server side rendering

Blasz:
Closes #323.

I've abstracted the reset method solution from #337 into separate server + style context functions to make it a little bit more future proof.

I've added both integration and unit tests for the new functionality and fixed the context from being shared between unit tests by calling these functions in `beforeEach`.


README.md
```diff
@@ -1250,6 +1250,22 @@ It is possible for your `Draggable` to contain interactive elements. By default
 
 You can opt out of this behavior by adding the `disableInteractiveElementBlocking` prop to a `Draggable`. However, it is questionable as to whether you should be doing so because it will render the interactive element unusable. If you need to *conditionally* block dragging from interactive elements you can add the `disableInteractiveElementBlocking` prop to opt out of the default blocking and monkey patch the `dragHandleProps (DragHandleProps)` event handlers to disable dragging as required.
 
+## `resetServerContext`
```
alexreardon: Do you think we should make a new section 'Server side rendering' with `resetServerContext` underneath?

Blasz: I was considering this as well but ultimately opted to keep the heading as the API method to be consistent with our other API exports. I think either way is fine though, up to you.

alexreardon: Agreed - let's leave it

alexreardon:
pingu2

PR #346, opened 2018-02-25
Reduce umd bundle size

TrySound:
before
react-beautiful-dnd.js 381.3K
react-beautiful-dnd.min.js 148.0K

with esm
react-beautiful-dnd.js 325.6K
react-beautiful-dnd.min.js 123.9K

with esm and disabled polyfills
react-beautiful-dnd.js 324.6K
react-beautiful-dnd.min.js 123.7K

with esm and embedded helpers
react-beautiful-dnd.js 285.1K
react-beautiful-dnd.min.js 108.4K

with esm, embedded helpers and loose mode
react-beautiful-dnd.js 283.3K
react-beautiful-dnd.min.js 107.5K

with esm, embedded helpers, loose mode and enabled polyfills
react-beautiful-dnd.js 294.4K
react-beautiful-dnd.min.js 112.0K

with esm, embedded helpers, loose mode and object-assign polyfill
react-beautiful-dnd.js 283.3K
react-beautiful-dnd.min.js 107.8K


package.json
```diff
@@ -27,8 +27,8 @@
     "typecheck": "flow check",
     "build": "yarn run build:clean && yarn run build:lib && yarn run build:flow && yarn run build:dist",
     "build:clean": "rimraf lib && rimraf dist",
-    "build:dist": "rollup -c && rollup -c --min",
-    "build:lib": "babel src -d lib",
+    "build:dist": "rollup -c && cross-env NODE_ENV=min rollup -c",
```
alexreardon: can you please explain your thinking behind this one?

TrySound: `--min` can be considered as rollup option. But it's not. This may confuse. Specifying env I think safer way to customize config.

alexreardon:
I am unfamiliar with the details of rollup. Can you please explain how these changes resulted in the big kb drop?

lukebatchelor:
I'm assuming the labels are mixed up above?"with helpers" should be larger right?

TrySound:
@lukebatchelor helpers are added by runtime

TrySound:
1. cjs is not compact in rollup. It's converted to esm with wrapper which adds a lot of bytes around code. I just transpile to esm via babel (disabled modules-commonjs).
2. babel-runtime helpers are commonjs and have the same story. external helpers plugin delegates helpers to babel-core which passes them to `rollup-plugin-babel` which adds them as esm.
3. babel-runtime adds polyfills with helpers.

lukebatchelor:
Also,what do you mean by "with esm"? This is still a cjs build right?

lukebatchelor:
Gotchya, answered above. Cheers

lukebatchelor:
Sweet, just checking my understanding here:

External helpers is used to make sure we don't get repeated instances of Babel helper functions (like class callbacks) which is likely a large part of the space saving here?
https://github.com/rollup/rollup-plugin-babel/blob/master/README.md

You're also running rollup on the src itself, rather than the CJ's because the CJ's has had the imports removed already and rollup would have d had to add them all back in and add wrapper code to make it work? (Presumably was being done by https://github.com/rollup/rollup-plugin-commonjs previously, and now we only use that on nodemodules?).

The numbers you gave at the top were the original size, then the size just by doing the transforms in a smarter order, then the size by using the modules versions of Babel runtime?


Looks great to me, thanks heaps for the contribution.

Alex, would you be comfortable doing a named alpha release of this so we can cross browser and integration test this?

lukebatchelor:
Apologies for all the autocorrected junk in there. I'm on mobile

TrySound:
@lukebatchelor Not quite. external-helpers and babel-runtime do similar thing. But babel-runtime is cjs. rollup-plugin-babel adds helpers as esm. Also rollup do not adds polyfills, which are enabled by default in babel-runtime.

TrySound:
I disabled runtime polyfills and added stats in topic.

.babelrc
```diff
@@ -2,26 +2,26 @@
...
     "transform-export-extensions"
   ],
-  "comments": false
-}
\ No newline at end of file
+  "comments": false,
+  "env": {
+    "cjs": {
+      "plugins": [
+        ["transform-runtime", { "polyfill": false }],
```
alexreardon: Should we be doing this?

package.json
```diff
@@ -51,13 +51,16 @@
     "@storybook/react": "^3.2.16",
     "babel-cli": "^6.26.0",
     "babel-eslint": "^8.2.1",
+    "babel-plugin-external-helpers": "^6.22.0",
     "babel-plugin-transform-class-properties": "^6.24.1",
+    "babel-plugin-transform-es2015-modules-commonjs": "^6.26.0",
```
alexreardon: What is the role of this one?

.babelrc
TrySound: runtime adds this polyfills
- babel-runtime/core-js/object/assign
- babel-runtime/core-js/object/get-prototype-of
- babel-runtime/core-js/object/keys
the case is that rollup do not adds any polyfill.

lukebatchelor: Won't this affect consumers who are using the cjs build though?

TrySound: Depends on browser support. I just get rid from get-prototype-of polyfill with loose mode.
Assign can be removed with spread operator which reuses extend helpers which is already in bundle. Not sure about `Object.keys`, but it's even es5.

TrySound: I see, it's only one place with mutating. Maybe local polyfill?

TrySound:
@Andarist did I miss something?

.babelrc
alexreardon: We'll have to check on ie11. I think it safest to leave the polyfills in for now given they are so small

alexreardon:
@lukebatchelor 

> Alex, would you be comfortable doing a named alpha release of this so we can cross browser and integration test this?

Yep

.babelrc
alexreardon: Thoughts @TrySound ?

TrySound: I updated stats. It seems like in polyfills has bigger impact external-helpers and loose mode. Not sure. I just added `object-assign` polyfill, which is small and the same as babel adds.

alexreardon: I think for the sake of 3kb it is worth keeping the babel polyfills for simplicity and for safety

alexreardon: 112.0K vs 107.8K (4.2kb)

alexreardon: I like the idea of keeping the babel one also as we go forward as it can add things we use dynamically

alexreardon: @lukebatchelor ^

TrySound: The problem is that babel is not actually good with helpers. It can detect only static stuff like `Object.keys`, but cannot `[].reduce`. So you still need to check by yourself. With current configuration runtime adds only `Object.keys` polyfill, which is useless.

lukebatchelor: Hmmm... I do agree it's kind dvof useless, but my worry would be someone adding in a usage later which would then break ie11.

There really aren't many things it helps with, but might be worth keeping just for the safety? Unless there's a simple eslint rule we can turn on to avoid it?

alexreardon: It would make me feel better using the babel auto polyfill incase we lean on it in the future. I know that what you have is strictly smaller - but i think the safety of using babel is worth keeping it for now.

If you can update the PR @TrySound to include that change we should be good to go!

Thanks for your perseverance with this one @TrySound !

TrySound:
Okay. We end up with
react-beautiful-dnd.js 317.7K
react-beautiful-dnd.min.js 121.1K

I can try more with babel beta. It has better options.

rollup.config.js
```diff
@@ -6,30 +6,35 @@ import uglify from 'rollup-plugin-uglify';
...
     name: 'ReactBeautifulDnd',
     globals: { react: 'React' },
   },
   plugins: [
-    babel({ exclude: 'node_modules/**', babelrc: false }),
-    resolve(),
-    commonjs(),
+    babel({
+      exclude: 'node_modules/**',
+      runtimeHelpers: true,
```
Andarist: why runtimeHelpers and not external ones? where external is ofc a misleading name here, because they would be included in the bundle

TrySound: It was. But guys want to keep polyfills.

alexreardon: For future safety: rather than adding polyfills we can delegate to babel

alexreardon: But if you think that is a terrible idea I am open to being convinced!

Andarist:
@TrySound 

Looks good, I also prefer cherry-picked polyfills (or ponyfills actually), but in the grand scheme of things - it's kinda whatever.

BTW are those stats for min.js given after gzip or before it?

TrySound:
Before. Just uglify.

alexreardon:
@TrySound is there anything else you wanted to try with this one or are you happy to move forward as is?

TrySound:
@alexreardon I'm okay with current state. It's better than before. We can investigate more in the next PR.

TrySound:
Just interesting how babel `comments: false` affects code?

alexreardon:
My understanding is that it drops all the comments (reduces the file size)

alexreardon:
I will try to get this merged tomorrow

alexreardon:
Thanks @TrySound, @lukebatchelor and @Andarist for your help with this! A really good outcome

Issue #10, opened 2017-08-10
Multi drag support


alexreardon:
Support the dragging of multiple Draggables. There are still a lot of open questions about this one. 

- Currently the library has no concept of *selecting* a `Draggable`. What would this look like to support 'select' - is this handled by the library - or is it a pattern?
- What is the expected visual experience for a multi drag
- What would the api for `onDragStart` and `onDragEnd` look like?

alexreardon:
We will be starting work on this soon 💖

souporserious:
Excited to see this is on the radar! Have you seen Google drive's multi drag interactions? I really love how they feel native when using them.

![multi-dnd](https://user-images.githubusercontent.com/2762082/37180712-e0e6d4a0-22de-11e8-8254-1e51fbd655ae.gif)

Maybe react-beautiful-dnd could provide some extra methods/components to achieve something similar? It would be awesome to somehow grab a bunch of nodes and have them animate towards the cursor.

alexreardon:
This has landed :) #383

dileepkeeppalli:
Can we have a documentation to use this feature?

alexreardon:
Here it is: https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/patterns/multi-drag.md

doubleedev:
Has anyone had any joy with getting this to work in react functional components (no typescript)?

PR #353, opened 2018-02-27
correctly blocking onClick for all element types

### LINKED ISSUE #273: Containing element receives mouse click after dragging
## Bug or feature request?
Bug

### Expected behavior
A mouse click event is not emitted on the parent element of the draggables after dragging and dropping.

### Actual behavior
A mouse click event *is* emitted on the parent element of the draggables after dragging and dropping.

### Browser version
Chrome Version 63.0.3239.84 (Official Build) (64-bit)

### Demo

See logs of the following demo after dragging and dropping:
https://www.webpackbin.com/bins/-L2yYhx5eZi1MJnlU-79

alexreardon:
Fixes #273.

This is required for #10. 

`onClick` event handlers are not firing on the `drag handle` because it has `pointer-events: none;` applied to it. This means that it fires on the parent. Confusingly the `onClick` handler will fire on the element if the element is an `anchor`. There is a work around (see #10) but it is hacky. This change would make `onClick` fire correctly on the `drag handle` every time. It involves removing `pointer-events: none` off the dragging `drag handle`. This has no performance implications.

There is a drawback. By changing this you will no longer be able to use the mouse wheel to scroll a `Droppable` underneath a dragging item. Disabling pointer events allowed the user to 'scroll through' the dragging item. Turning them back on means that they can no longer do this. However, I think given that users can now automatically scroll the window and scroll containers by using mouse movement this is acceptable. I have investigated the other drag and drop libraries out there that I could and they all have this restriction.

I think on the whole this is a positive and useful change as it delivers more thoroughly on our attempt to correctly block click events


src/view/style-marshal/get-styles.js
```diff
@@ -113,8 +114,8 @@ export default (styleContext: string): Styles => {
   // > Applied while the user is actively dragging
 
   // cursor: grab
-  // We apply this by default for an improved user experience. It is such a common default that we
-  // bake it right in. Consumers can opt out of this by adding a selector with higher specificity
+  // We apply apply this style to the body in case the user pointer
```
jaredcrowe: absolute nitpick but you've written 'apply' here twice

alexreardon: i'll fix it

alexreardon: fixed

alexreardon:
It sucks we need to do this - but it is for the best

alexreardon:
I was not happy to loose wheel scrolling so I have changed how we do post drag click prevention. It is working really well.

alexreardon:
A point for disucssion is whether this should be a breaking change. I have currently removed `onClick` from `DragHandleProps` which is a breaking change. I add it back in and make it a no-opt to make this a non breaking change. However, doing this means that future consumers need to monkey patch onClick for no reason until we do the next major release (which sucks).

I am not sure what the right path forward is. Thoughts @jaredcrowe @Blasz ?

My personal thought is to keep the api as tight as possible and drop `onClick` and move to `6.0`.

README.md
```diff
@@ -366,6 +366,12 @@ point-events: none;
 
 **Styles applied using the `data-react-beautiful-dnd-draggable` attribute**
 
+We apply a cursor while dragging to give user feedback that a drag is occurring. You are welcome to override this by applying your own cursor to the element with a greater specificity.
```
alexreardon: I need to remove these new comments (they are no longer needed)

README.md
```diff
@@ -1142,23 +1146,20 @@ Controlling a whole draggable by just a part of it
 You can override some of the `dragHandleProps` props with your own behavior if you need to.
 
 ```js
-const myOnClick = event => console.log('clicked on', event.target);
+const myOnMouseDown = event => console.log('mouse down on', event.target);
```
alexreardon: no longer using 'click' as part of the example

src/view/style-marshal/get-styles.js
```diff
@@ -65,9 +64,11 @@ export default (styleContext: string): Styles => {
         cursor: grab;
       }
     `,
-    blockPointerEvents: `
+    whileActiveDragging: `
```
alexreardon: This needs to be undone

alexreardon:
I have tested this on our supported desktop browsers. I still need to test it on mobile (ios and andriod)

alexreardon:
It is looking good on ios. I'll check andriod after lunch

jaredcrowe:
If it wasn't for the fact that anchors won't always get their onClick called now I would have said make DragHandleProps.onClick a no-op and remove it next major. But since that behaviour is changing I agree that we should probably make a breaking change.

PR #356, opened 2018-02-28
Using box model terms

### LINKED ISSUE #355: Improve internal box model terminology
Currently we use the following language to describe different areas associated with a dimension:

- `withoutMargin`: straight from `getBoundingClientRect`
- `withMargin`: we use `window.getComputedStyles(el)` to pull the margin from the element it and add it to `withoutMargin`
- `withoutPadding`: we use `window.getComputedStyles(el)` to obtain the elements padding and then we shrink the `withoutMargin` dimension by this amount.

These terms have grown organically as needs arose. They also directly map to the [box model](https://developer.mozilla.org/en-US/docs/Learn/CSS/Introduction_to_CSS/Box_model). To reduce friction we should rename our properties to match their box model equivalents

- `withMargin` -> `marginBox`
- `withoutMargin` -> `paddingBox`
- `withoutPadding` -> `contentBox`

For now we are getting away with ignoring borders 😊 but we could look at adding a `borderBox` as well. Although, we could do this as a seperate task

alexreardon:
Fixes #355


src/state/dimension.js
```diff
@@ -159,24 +157,32 @@ type GetDroppableArgs = {|
...
   direction = 'vertical',
   margin = noSpacing,
   padding = noSpacing,
   windowScroll = origin,
   isEnabled = true,
 }: GetDroppableArgs): DroppableDimension => {
-  const withMargin: Spacing = expandBySpacing(client, margin);
-  const withWindowScroll: Spacing = offsetByPosition(client, windowScroll);
-  const subject: Area = getArea(expandBySpacing(withWindowScroll, margin));
+  const client: BoxModel = {
```
alexreardon: so much cleaner!

src/types.js
```diff
@@ -88,13 +85,13 @@ export type DraggableDimension = {|
   placeholder: Placeholder,
   // relative to the viewport when the drag started
   client: {|
```
alexreardon: not adding conentBox to Draggables yet as it is not used

src/types.js
```diff
@@ -130,24 +127,42 @@ export type DroppableDimensionViewport = {|
   clipped: ?Area,
 |}
 
+// Maps to the CSS box model
```
alexreardon: Love this!

src/state/dimension.js
```diff
@@ -28,37 +29,34 @@ export const noSpacing: Spacing = {
 
 type GetDraggableArgs = {|
   descriptor: DraggableDescriptor,
-  client: Area,
+  paddingBox: Area,
```
alexreardon: renamed this to be clearer: otherwise, 'what is a client?'

src/state/dimension.js
```diff
@@ -28,37 +29,34 @@ export const noSpacing: Spacing = {
...
   margin?: Spacing,
   windowScroll?: Position,
 |};
 
 export const getDraggableDimension = ({
   descriptor,
-  client,
+  paddingBox,
```
alexreardon: renamed param

src/state/dimension.js
```diff
@@ -139,10 +137,10 @@ export const scrollDroppable = (
 
 type GetDroppableArgs = {|
   descriptor: DroppableDescriptor,
-  client: Area,
+  paddingBox: Area,
   // optionally provided - and can also be null
   closest?: ?{|
-    frameClient: Area,
+    framePaddingBox: Area,
```
alexreardon: renamed for clarity

src/state/dimension.js
```diff
@@ -139,10 +137,10 @@ export const scrollDroppable = (
 
 type GetDroppableArgs = {|
   descriptor: DroppableDescriptor,
-  client: Area,
+  paddingBox: Area,
```
alexreardon: renamed for clarity

src/state/dimension.js
```diff
@@ -159,24 +157,32 @@ type GetDroppableArgs = {|
 
 export const getDroppableDimension = ({
   descriptor,
-  client,
+  paddingBox,
```
alexreardon: renamed :)

src/state/dimension.js
```diff
@@ -159,24 +157,32 @@ type GetDroppableArgs = {|
...
 }: GetDroppableArgs): DroppableDimension => {
-  const withMargin: Spacing = expandBySpacing(client, margin);
-  const withWindowScroll: Spacing = offsetByPosition(client, windowScroll);
-  const subject: Area = getArea(expandBySpacing(withWindowScroll, margin));
+  const client: BoxModel = {
+    paddingBox,
+    marginBox: getArea(expandBySpacing(paddingBox, margin)),
+    contentBox: getArea(shrinkBySpacing(paddingBox, padding)),
+  };
+  const page: BoxModel = {
```
alexreardon: so simple!

src/types.js
```diff
@@ -75,10 +75,7 @@ export type Axis = VerticalAxis | HorizontalAxis
 export type Placeholder = {|
   // We apply the margin separately to maintain margin collapsing
   // behavior of the original element
-  withoutMargin: {|
-    height: number,
-    width: number,
-  |},
+  paddingBox: Area,
```
alexreardon: rather than just passing a cut I just pass the whole thing now. Because we name it `paddingBox` it is clearer that we are passing the right thing

PR #358, opened 2018-03-02
less dom reads

### LINKED ISSUE #329: Move viewport to state
Currently we call `getViewport` all over the place to get the current viewport. This was done for simplicity and because when keyboard dragging we ignored window scrolling... 

This is a non-functional issue. It just feels wrong to have state functions reaching to the window to grab it's current state. Ideally we would just store the `viewport` in the state while dragging and update it when a window scroll occurs.

We do not need to worry about it changing dimensions as we cancel a drag on any resize.

alexreardon:
Closes #329.

Fixing a minor performance regressed in lifting big data sets with our new auto scroller.

What this PR does:

- avoids independent calls to functions that request values from the window from the `state` folder (engineering health)
- Fixes minor perf regression on lift. This was caused by reading window properties at bad times in an animation frame. There were a few ways around it. But this method gets around the problem while also doing something we wanted to anyway.

The code changes in `src` are super basic. There was some more heavy lifting required to get everything working in the tests. Previously tests were setting viewport properties in anticipation of the viewport being read. Now these properties can be passed directly to functions


src/state/auto-scroller/can-scroll.js
```diff
@@ -94,38 +91,12 @@ export const canPartiallyScroll = ({
...
-  if (!el) {
-    return origin;
-  }
-
-  const viewport: Area = getViewport();
-
-  // window.innerWidth / innerHeight includes scrollbar
-  // however the scrollHeight / scrollWidth do not :(
-
-  const maxScroll: Position = getMaxScroll({
```
alexreardon: this is now computed in getViewport

src/state/debug-middleware/timing-average-middleware.js
```diff
@@ -0,0 +1,46 @@
+// @flow
```
alexreardon: another debug middleware - this one averages the time taken for actions and prints a result

src/state/move-cross-axis/get-best-cross-axis-droppable.js
```diff
@@ -49,8 +51,6 @@ export default ({
     sourceClipped[axis.start],
     sourceClipped[axis.end]
   );
-  const viewport: Area = getViewport();
```
alexreardon: that's what i'm talking about 👍

src/state/move-to-next-index/in-foreign-list.js
```diff
@@ -74,7 +73,6 @@ export default ({
     return 'start';
   })();
 
-  const viewport: Area = getViewport();
```
alexreardon: goodbye old friend

src/state/reducer.js
```diff
@@ -478,22 +481,23 @@ export default (state: State = clean('IDLE'), action: Action): State => {
...
+    const { viewport } = action.payload;
     const drag: ?DragState = state.drag;
 
     if (!drag) {
       console.error('cannot move with window scrolling if no current drag');
       return clean();
     }
 
-    if (isEqual(windowScroll, drag.current.windowScroll)) {
+    // TODO: need to improve this so that it compares the whole viewport
```
alexreardon: I think this check is good enough for now

src/view/draggable/draggable.jsx
```diff
@@ -13,7 +13,7 @@ import type {
 import DraggableDimensionPublisher from '../draggable-dimension-publisher/';
 import Moveable from '../moveable/';
 import DragHandle from '../drag-handle';
-import getWindowScroll from '../../window/get-window-scroll';
+import getViewport from '../window/get-viewport';
```
alexreardon: this is now the only thing that publishes the viewport into the state

src/view/window/get-viewport.js
```diff
@@ -0,0 +1,41 @@
...
+  // Computed
+  const right: number = left + width;
+  const bottom: number = top + height;
+
+  const subject: Area = getArea({
+    top, left, right, bottom,
+  });
+
+  const maxScroll: Position = getMaxScroll({
+    scrollHeight: doc.scrollHeight,
```
alexreardon: It turns out these properties are super slow to read sometimes

stories/src/horizontal/author-app.jsx
```diff
@@ -49,6 +49,10 @@ export default class AuthorApp extends Component<Props, State> {
       return;
     }
 
+    if (result.destination.index === result.source.index) {
```
alexreardon: A little optimisation for our storybooks

test/unit/integration/lift-action-and-dimension-marshal.spec.js
```diff
@@ -30,7 +33,18 @@ import type {
 } from '../../../src/types';
 
 const preset = getPreset();
-const origin: Position = { x: 0, y: 0 };
+
+const scrolledViewport: Viewport = createViewport({
```
alexreardon: much clearer now

test/unit/state/auto-scroll/jump-scroller.spec.js
```diff
@@ -25,18 +24,22 @@ const windowScrollSize = {
...
+  scrollWidth: windowScrollSize.scrollWidth,
+  scroll: origin,
 });
 
-const disableWindowScroll = () => {
-  setWindowScrollSize({
-    scrollHeight: viewport.height,
-    scrollWidth: viewport.width,
-  });
+const unscrollableViewport: Viewport = {
```
alexreardon: yum

test/utils/viewport.js
```diff
@@ -0,0 +1,110 @@
+// @flow
```
alexreardon: improved helpers

PR #365, opened 2018-03-04
Render timing fix

### LINKED ISSUE #357: Animation (bug?) on Firefox
## Bug or feature request?
Bug may be ?

### Expected behavior

No re-render of the animation.

### Actual behavior & Steps to reproduce

On Firefox, when you right clic on an element and have it in selection and then move it down in the list, the elements placed before the new position have their animation re-render or something like this... Is this a normal behavior ?

[Demo](https://codesandbox.io/s/k260nyxq9v)

### Browser version

Firefox 58.0.2

### Demo

This is from your vertical list basic usage:

alexreardon:
Fixes #357

### DIFF
--- a/src/view/drag-drop-context/drag-drop-context.jsx
+++ b/src/view/drag-drop-context/drag-drop-context.jsx
@@ -150,31 +150,45 @@ export default class DragDropContext extends React.Component<Props> {
     let previous: State = this.store.getState();
 
     this.unsubscribe = this.store.subscribe(() => {
-      const previousValue: State = previous;
+      const previousInThisExecution: State = previous;
       const current = this.store.getState();
       // setting previous now incase any of the
       // functions synchronously trigger more updates
       previous = current;
 
-      // TODO: this probs needs to be done first
+      const isPhaseChanging: boolean = current.phase !== previousInThisExecution.phase;
+
+      // Style updates do not cause more actions. It is important to update styles
+      // before hooks are called: specifically the onDragEnd hook. We need to clear
+      // the transition styles off the elements before a reorder to prevent strange
+      // post drag animations in firefox. Even though we clear the transition straight
+      // after the reorder - firefox continues to apply the transition.
+      if (isPhaseChanging) {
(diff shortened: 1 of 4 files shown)


src/view/drag-drop-context/drag-drop-context.jsx
```diff
@@ -150,31 +150,46 @@ export default class DragDropContext extends React.Component<Props> {
     let previous: State = this.store.getState();
 
     this.unsubscribe = this.store.subscribe(() => {
-      const previousValue: State = previous;
       const current = this.store.getState();
```
alexreardon: The only change here was to move the style marshal update before the hooks are called. However, I wanted to add some explanation to what is going on because the timing of everything is really important

src/view/drag-drop-context/drag-drop-context.jsx
```diff
@@ -150,31 +150,46 @@ export default class DragDropContext extends React.Component<Props> {
...
-      // setting previous now incase any of the
-      // functions synchronously trigger more updates
+      const previousInThisExecution: State = previous;
```
alexreardon: Updated name to be clearer

src/view/drag-drop-context/drag-drop-context.jsx
```diff
@@ -150,31 +150,46 @@ export default class DragDropContext extends React.Component<Props> {
...
       previous = current;
 
-      // TODO: this probs needs to be done first
+      // Style updates do not cause more actions. It is important to update styles
+      // before hooks are called: specifically the onDragEnd hook. We need to clear
+      // the transition styles off the elements before a reorder to prevent strange
+      // post drag animations in firefox. Even though we clear the transition off
+      // a Draggable - if it is done after a reorder firefox will still apply the
+      // transition.
+      if (isPhaseChanging) {
```
alexreardon: style marshal now called first

alexreardon:
Thanks @jaredcrowe

PR #368, opened 2018-03-05
phrasing content within a button

### LINKED ISSUE #359: (Bug ?) Nested interactive elements : Button with other than only text
Firstly, thanks for this amazing work !

## Bug or feature request?
Nested interactive elements works very well. But, when you put something in the button, other than text, (`<strong></strong>` for example) dragging is no more blocked.

### Expected behavior
Dragging should be blocked no matter what button contains. Useful for svg icons for example.

### Actual behavior
I can drag an element if it's not only text in the button. 

### Steps to reproduce
Just put `<strong>Text</strong>` in a button.

### Browser version
Chrome 64.0.3282.186

### Doubt
Maybe I'm wrong and it's not a bug. So, is there a workaround ?

alexreardon:
Fixes #359


alexreardon:
I have run over these changes with @seancurtis and he is happy with them

src/view/drag-handle/util/should-allow-dragging-from-target.js
```diff
@@ -16,15 +16,26 @@ export const interactiveTagNames: TagNameMap = {
...
     return false;
   }
 
+  // Most interactive elements cannot have children. However, some can such as 'button'.
+  // See 'Permitted content' on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button
+  // Rather than having two different functions we can consolidate our checks into this single
+  // function to keep things simple.
+  // There is no harm checking if the parent is has an interactive tag name even if it cannot have
+  // any children. We need to perform this loop anyway to check for the contenteditable attribute
+  const hasAnInteractiveTag: boolean = Boolean(interactiveTagNames[current.tagName.toLowerCase()]);
```
Blasz: Any reason why you're using the `Boolean` constructer to coerce to boolean over `!!`? `!!` is more performant but if this isn't being called a lot then it won't make a difference here. Micro optimisations aren't worth it :) https://jsperf.com/bool-vs-doublenot

src/view/drag-handle/util/should-allow-dragging-from-target.js
```diff
@@ -16,15 +16,26 @@ export const interactiveTagNames: TagNameMap = {
...
+const isAnInteractiveElement = (parent: Element, current: ?Element): boolean => {
   if (current == null) {
     return false;
   }
 
+  // Most interactive elements cannot have children. However, some can such as 'button'.
+  // See 'Permitted content' on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button
+  // Rather than having two different functions we can consolidate our checks into this single
+  // function to keep things simple.
+  // There is no harm checking if the parent is has an interactive tag name even if it cannot have
```
Blasz: Typo here: 'parent is has'

test/unit/view/drag-handle.spec.js
```diff
@@ -2522,72 +2522,406 @@ describe('drag handle', () => {
...
+                expect(callbacksCalled(callbacks)({
+                  onLift: index + 1,
+                  onDrop: index + 1,
+                })).toBe(true);
+              });
+            });
+
+            it('should start a drag if the target is not an interactive element', () => {
+              const nonInteractiveTagNames: TagNameMap = {
+                a: true,
```
Blasz: Just out of interest, is there any reason why anchors are not classed as interactive elements?

src/view/drag-handle/util/should-allow-dragging-from-target.js
alexreardon: I think it is clearer

https://github.com/getify/You-Dont-Know-JS/blob/master/types%20%26%20grammar/ch4.md#explicitly-----boolean

alexreardon: It also returns a value (it would return an `Boolean` object if we used `new Boolean(value)`

alexreardon: Rather that using a double negative to cast - we are **super explicitly** casting to boolean :)

src/view/drag-handle/util/should-allow-dragging-from-target.js
alexreardon: i'll fix

stories/src/interactive-elements/interactive-elements-app.jsx
```diff
@@ -19,7 +19,20 @@ const initial: ItemType[] = [
...
+        <p>With child {' '}
+          <a
+            href="https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Content_categories#Phrasing_content"
+            target="_blank"
+            rel="noopener noreferrer"
+          >
+              phrasing content
+          </a>
+        </p>
+        <button>hello <strong>world</strong></button>
```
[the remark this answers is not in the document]
alexreardon: This is not a test file - it is simply a storybook. I have added tests for a parent, but not multi levels of parentage. I am not sure if a test would be super valueable - thoughts?

test/unit/view/drag-handle.spec.js
alexreardon: Their only interaction is a click. We actually allow draggables and drag handles to be anchors - so they can be both clickable as well as draggable. We manage this interaction separately. We completely support an anchor being an anchor as well as draggable. This feels right too. The other elements are more form elements and do not make sense as a thing if you cannot interact with it directly.

Does that help?

Issue #235, opened 2017-12-21
ES module build


foisonocean:
## Bug or feature request?
feature request

Could you provide es module build, which make tree-shaking available in build tools like `webpack` and `rollup`?

alexreardon:
We currently have a distribution build, but an esmodule one would be useful as well.

![21i2jk](https://user-images.githubusercontent.com/2182637/34277213-ea53a260-e6f8-11e7-903e-2f03f80edd0f.jpg)

alexreardon:
Closed by #369

PR #369, opened 2018-03-05
Build ES6 module

b1f6c1c4:
For those who prefer ES6 import/export (such as rollup and webpack user), it'd be better not to babel `transform-es2015-modules-commonjs`.


package.json
```diff
@@ -16,6 +16,8 @@
     "url": "https://github.com/atlassian/react-beautiful-dnd/issues"
   },
   "main": "lib/index.js",
+  "module": "lib/es/index.js",
```
TrySound: Can you call it `esm`?

package.json
```diff
@@ -26,8 +28,9 @@
     "validate": "yarn run lint && yarn run typecheck",
     "lint": "eslint .",
     "typecheck": "flow check",
-    "build": "yarn run build:clean && yarn run build:lib && yarn run build:flow && yarn run build:dist",
+    "build": "yarn run build:clean && yarn run build:lib && yarn run build:flow && yarn run build:dist && yarn run build:es",
     "build:clean": "rimraf lib && rimraf dist",
+    "build:es": "cross-env NODE_ENV=production BABEL_ENV=es babel ./src --out-dir ./lib/es",
```
TrySound: `esm` and babel_env is not required here

.babelrc
```diff
@@ -12,6 +12,10 @@
   ],
   "comments": false,
   "env": {
+    "es": {
```
TrySound: this is not required

b1f6c1c4:
@TrySound BTW, should we assume that those 'lib/es' users are able to accept ES6 so `transform-runtime`, `transform-class-properties`, `transform-object-rest-spread`, `transform-export-extensions` are unnecessary for them?

TrySound:
@b1f6c1c4 Nope. That's why I ask you to call it `esm`.

package.json
b1f6c1c4: To my knowledge babel uses `BABEL_ENV` to match the `env` section in `.babelrc`. It's possible to rename it here to `BABEL_ENV=esm` (and modify in `.babelrc`) but it won't work if we remove `BABEL_ENV`.

TrySound: I meant rename task to esm and remove BABEL_ENV which is not used.

TrySound: Btw production is not required here too

b1f6c1c4:
@TrySound Just a kind reminder that in the 1027 packages used by my project, `es` is much more popular than `esm`. And (to my knowledge) they use ES5 syntax with ES6 import/exports.
```
history/es
i18next/dist/es
i18next-browser-languagedetector/dist/es
jquery-i18next/dist/es
material-ui/es
react-beautiful-dnd/lib/es
react-redux/es
react-router/es
react-router-dom/es
react-router-redux/es
recompose/es
redux/es
redux-form/es
redux-form-material-ui/es
redux-saga/es
redux-thunk/es
reselect/es
symbol-observable/es

acorn/dist/acorn_loose.es.js
acorn/dist/walk.es.js
acorn/dist/acorn.es.js
acorn/dist/acorn_loose.es.js
acorn/dist/walk.es.js
brcast/dist/brcast.es.js
deep-diff/index.es.js
immutable/dist/immutable.es.js
material-ui/index.es.js
material-ui-icons/index.es.js
node-fetch/lib/index.es.js
react-intl/lib/index.es.js
react-side-effect/lib/index.es.js

popper.js/dist/esm
theming/dist/esm
```

TrySound:
It doesn't matter what "they" use. From users feedback I got that `es` often confuses. So it's much better to rely [@std/esm](https://github.com/standard-things/esm) name.

And you asked about removing `transform-runtime`, `transform-class-properties`, `transform-object-rest-spread`, `transform-export-extensions`, but they are converts to what any users able to consume.

package.json
TrySound: also I'd place it in `esm/index.js`

package.json
```diff
@@ -16,6 +16,8 @@
     "url": "https://github.com/atlassian/react-beautiful-dnd/issues"
   },
   "main": "lib/index.js",
+  "module": "esm/index.js",
+  "sideEffects": false,
   "files": [
     "/lib",
```
TrySound: also add esm here. It's npm filter

TrySound:
ignore esm here https://github.com/atlassian/react-beautiful-dnd/blob/master/.eslintignore

alexreardon:
This looks good to me! Thanks @b1f6c1c4 for raising this and seeing it through. Also thanks @TrySound for jumping on this one and providing a great review.

dance-3

lukebatchelor:
Sweet, this looks awesome. This closes #235 (which has been sitting assigned to me for a while 😨).

Thanks for the input @TrySound, I had the same question about whether `es` should be "es5 with es6 imports" and whether you should be transpiling away classes, spread, etc. So thanks for clearing that up, TIL.

alexreardon:
This will go out in 6.0 which will be shipping soon

PR #379, opened 2018-03-09
Improved event handling

### LINKED ISSUE #375: Improve how raw input events are handled
Right now we put a [fair amount of effort](https://github.com/atlassian/react-beautiful-dnd#sloppy-clicks-and-click-blocking-) into preventing click events from being published after a drag.

As a part of implementing a muti-drag pattern (#10) we need to be consistent with how we are processing events (`mousedown`, `mouseup`, `click`, and so on). Right now we sometimes block these events (using `event.stopPropogation()`) and prevent their default behaviour (using `event.preventDefault()`), sometimes let them pass through. This can result in unmatched event pairs.

For example, in a mouse drag we stop the mousedown event being propogated. However, if they do not start a drag we allow the mouseup event to go through. The result of this is that consumers can get a mouseup without a proceeding mousedown event which can be confusing. In that situation we could also block the mousedown even if the user was not dragging but then a click will be published which might be unexpected given that there was no previous mouseup and mousedown event. There are other scenarios where we block on events in a pair (eg only block one of keydown and keyup).

### Proposed solution

Rather than conditionally stopping events with event.stopPropagation() we should publish every event and only use event.preventDefault() on events we are controlling. This way consumers can get every event as well as knowing if they should use it or not through the event.defaultPrevented property.

alexreardon:
Closes #375


src/view/drag-handle/util/bind-events.js
```diff
@@ -0,0 +1,35 @@
...
+
+export const unbindEvents = (
+  el: HTMLElement,
+  bindings: EventBinding[],
+  sharedOptions?: EventOptions,
+) => {
+  bindings.forEach((binding: EventBinding) => {
+    const options: Object = {
+      ...sharedOptions,
+      ...binding.options,
```
alexreardon: bindings can have their own overrides

src/debug/timings.js
```diff
@@ -0,0 +1,48 @@
...
+type Records = {
+  [key: string]: number,
+}
+
+const records: Records = {};
+
+const flag: string = '__react-beautiful-dnd-debug-hook__';
+
+// temp
+// window[flag] = true;
```
Blasz: Should this be left in?

src/view/drag-handle/drag-handle-types.js
```diff
@@ -36,8 +36,7 @@ export type DragHandleProps = {|
 
   // Stop html5 drag and drop
   draggable: boolean,
-  onDragStart: () => boolean,
-  onDrop: () => boolean
```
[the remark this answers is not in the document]
Blasz: Should all references to `onDrop` be removed or do they refer to different things? Namely the `onDrop` prop in the `Callbacks` type above.

src/view/drag-handle/sensor/create-keyboard-sensor.js
```diff
@@ -161,53 +171,65 @@ export default ({
...
-
-    // blocking scroll jumping at this time
+    // preventing scroll jumping at this time
     if (scrollJumpKeys[event.keyCode]) {
-      stopEvent(event);
+      event.preventDefault();
+      return;
     }
+
+    preventStandardKeyEvents(event);
```
Blasz: Would it be worthwhile adding the keycodes that we're specifically handling above into this function as well so that we don't need to add `event.preventDefault` inside each conditional?

src/view/drag-handle/sensor/create-mouse-sensor.js
```diff
@@ -41,7 +46,8 @@ export default ({
   const isDragging = (): boolean => state.isDragging;
   const isCapturing = (): boolean => Boolean(state.pending || state.isDragging);
   const schedule = createScheduler(callbacks);
-  const clickBlocker: ClickBlocker = createClickBlocker();
+  const getWindow = (): HTMLElement => getWindowFromRef(getDraggableRef());
+  const postDragEventPreveter: EventPreventer = createPostDragEventPreventer(getWindow);
```
Blasz: Variable typo

test/unit/view/drag-handle.spec.js
```diff
@@ -353,45 +413,84 @@ describe('drag handle', () => {
...
-        describe('cancelled with escape', () => {
-          beforeEach(() => {
-            mouseDown(wrapper, origin, auxiliaryButton);
-            // not moved enough yet
-            windowMouseMove({ x: 0, y: sloppyClickThreshold - 1 });
-            windowEscape();
+        describe('cancelled with any keydown', () => {
+          Object.keys(keyCodes).forEach((key: string) => {
+            describe(`with the ${key} key`, () => {
+              it('should not call execute any callbacks', () => {
```
Blasz: Typo I think, one of call/execute should be removed?

test/unit/view/drag-handle.spec.js
```diff
@@ -820,9 +958,39 @@ describe('drag handle', () => {
         })).toBe(true);
       });
 
-      it('should not do anything if there is nothing dragging', () => {
+      it('should block a click after a cancel', () => {
```
Blasz: block -> prevent?

test/unit/view/drag-handle.spec.js
```diff
@@ -1,5 +1,6 @@
 // @flow
```
Blasz: This is a big file!

src/debug/timings.js
alexreardon: I'll clean it up

test/unit/view/drag-handle.spec.js
alexreardon: yep. needs to be split up

src/view/drag-handle/sensor/create-mouse-sensor.js
alexreardon: fixed

test/unit/view/drag-handle.spec.js
alexreardon: fixed

test/unit/view/drag-handle.spec.js
alexreardon: yep. i have also updated the other places that used the blocked language

src/view/drag-handle/sensor/create-keyboard-sensor.js
alexreardon: these keys are ones that are prevented by standard across multiple sensors

alexreardon: the keyboard sensor has its own unique keyboard controls

README.md
```diff
@@ -1256,7 +1262,7 @@ The `children` function is also provided with a small amount of state relating t
 
 ### Adding an `onClick` handler to a `Draggable` or a *drag handle*
 
-You are welcome to add your own `onClick` handler to a `Draggable` or a *drag handle* (which might be the same element). `onClick` events handlers will only be called if we do not block the click. We block click events from occurring when the user was dragging an item. See [#sloppy-clicks-and-click-blocking-](sloppy clicks and click blocking) for more information.
+You are welcome to add your own `onClick` handler to a `Draggable` or a *drag handle* (which might be the same element). `onClick` events handlers will only be called if we do not block the click. We block click events from occurring when the user was dragging an item. See [#sloppy-clicks-and-click-prevention-](sloppy clicks and click prevention) for more information.
```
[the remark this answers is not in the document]
alexreardon: correct!

PR #383, opened 2018-03-12
Adding a multi drag pattern

alexreardon:


stories/src/multi-drag/reorder.old.js
```diff
@@ -0,0 +1,232 @@
+// @flow
```
Blasz: I can't see this file being imported anywhere, is it being used?

stories/src/multi-drag/reorder.old.js
```diff
@@ -0,0 +1,232 @@
...
+  columns,
+  selected,
+  source,
+  destination,
+}): Result => {
+  // 1. remove all of the selected tasks from their lists
+  // When ordering the collected tasks:
+  //  dragged item first
+  //  followed by the items with the lowest index
+  //  in the event of a tie, use the one that was selected first
```
Blasz: Should this order comment be located closer to the `collection.sort` line?

stories/src/multi-drag/reorder.old.js
```diff
@@ -0,0 +1,232 @@
...
+  tasks,
+});
+
+const reorderMultiDrag = ({
+  columns,
+  selected,
+  source,
+  destination,
+}): Result => {
+  // 1. remove all of the selected tasks from their lists
```
Blasz: This is the only numbered comment in the function. Should we remove the number or add some to the other comments?

stories/src/multi-drag/utils.js
```diff
@@ -0,0 +1,266 @@
...
+    columns: withAddedTasks,
+  };
+
+  return {
+    entities: updated,
+    selectedTaskIds: orderedSelectedTaskIds,
+  };
+};
+
+export const reorderBetweenColumns = (args: Args): Result => {
```
Blasz: Doesn't this function reorder tasks in the same column as well? Is there a better name for this?

alexreardon: Hmm!

alexreardon: `mutliDragAwareReorder` seems alright for now

LuoweiC:
I want to use this,but i can't use，can you give me some guidance?

Frankcarvajal:
@LuoweiC What specifically about it can you not use. May you provide some specific issues you are dealing with?

I've been able to recreate a multi drag supported task list in my own open source repository by using the src code story from here if that may assist I can link it.

geobde:
@Frankcarvajal can you provide an example?

Frankcarvajal:
@geobde I've got an example repository on my GitHub, but it doesn't have an example of a multi drag and drop table, just standard like an example. @alexreardon has no multi drag and drop table example and the issue I put in for this assigned to me, I've not had a moment to work on. I do wish he would put an example of it up

PR #386, opened 2018-03-15
Upgrade flow

TrySound:
- upgraded flow
- upgraded definitions with `flow-typed install --skip`
- fixed opaque types of `requestAnimationFrame` and `setTimeout`
- removed $ExpectError suppressions which may not exists in user config
- removed $FlowFixMe suppressions because sometimes they force flow recheck infinitely (personal experience)
- fixed a few suppressed errors
- prevented react-redux errors with more safe `any`, I think it's ok since you annotate so much

P.S. New react-redux definitions are still broken, probably because they do not cover all cases

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -12,6 +12,11 @@
     "node": true,
     "jest/globals": true
   },
+  "globals": {
+    "TimeoutID": true,
+    "IntervalID": true,
+    "AnimationFrameID": true
+  },
   // custom rules
   "rules": {
     // http://eslint.org/docs/rules/comma-dangle
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "eslint-plugin-jest": "^21.6.1",
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-react": "^7.5.1",
-    "flow-bin": "0.61.0",
+    "flow-bin": "0.67.1",
     "jest": "^22.0.5",
     "raf-stub": "^2.0.1",
     "react": "^15.6.2",
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "eslint-plugin-jest": "^21.6.1",
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-react": "^7.5.1",
-    "flow-bin": "0.67.1",
+    "flow-bin": "0.68.0",
     "jest": "^22.0.5",
     "raf-stub": "^2.0.1",
     "react": "^15.6.2",
--- a/src/state/get-new-home-client-center.js
+++ b/src/state/get-new-home-client-center.js
@@ -71,7 +71,6 @@ export default ({
     }
 
     // Otherwise, return the dimension of the empty foreign droppable
-    // $ExpectError - flow does not correctly type this as non optional
     return destination.client.contentBox;
   })();
 
--- a/src/state/reducer.js
+++ b/src/state/reducer.js
@@ -266,14 +266,13 @@ export default (state: State = clean('IDLE'), action: Action): State => {
         return existing;
       }
 
-      // $ExpectError - using spread
-      const newDrag: DragState = {
+      const newDrag: DragState = ({
         ...existing,
         current: {
           ...existing.current,
           hasCompletedFirstBulkPublish: true,
         },
-      };
+      }: any);
 
       return newDrag;
     })();
--- a/src/view/drag-handle/sensor/create-touch-sensor.js
+++ b/src/view/drag-handle/sensor/create-touch-sensor.js
@@ -15,7 +15,7 @@ import type { TouchSensor, CreateSensorArgs } from './sensor-types';
 type State = {
   isDragging: boolean,
   hasMoved: boolean,
-  longPressTimerId: ?number,
+  longPressTimerId: ?TimeoutID,
   pending: ?Position,
 }
 
@@ -96,7 +96,7 @@ export default ({
       y: clientY,
     };
 
-    const longPressTimerId: number = setTimeout(startDragging, timeForLongPress);
+    const longPressTimerId: TimeoutID = setTimeout(startDragging, timeForLongPress);
 
     setState({
       longPressTimerId,
@@ -108,7 +108,9 @@ export default ({
   };
 
   const stopPendingDrag = () => {
-    clearTimeout(state.longPressTimerId);
+    if (state.longPressTimerId) {
+      clearTimeout(state.longPressTimerId);
+    }
     schedule.cancel();
     touchStartMarshal.reset();
     unbindWindowEvents();
--- a/src/view/draggable/connected-draggable.js
+++ b/src/view/draggable/connected-draggable.js
@@ -1,4 +1,5 @@
 // @flow
+import type { Node } from 'react';
 import memoizeOne from 'memoize-one';
 import { connect } from 'react-redux';
 import { createSelector } from 'reselect';
@@ -32,6 +33,7 @@ import type {
 import type {
   MapProps,
   OwnProps,
+  DefaultProps,
   DispatchProps,
   Selector,
 } from './draggable-types';
@@ -262,12 +264,19 @@ const mapDispatchToProps: DispatchProps = {
 // Leaning heavily on the default shallow equality checking
 // that `connect` provides.
 // It avoids needing to do it own within `Draggable`
-export default connect(
+const ConnectedDraggable: OwnProps => Node = (connect(
   // returning a function to ensure each
   // Draggable gets its own selector
   (makeMapStateToProps: any),
-  mapDispatchToProps,
+  (mapDispatchToProps: any),
   null,
   { storeKey },
-)(Draggable);
+): any)(Draggable);
 
+ConnectedDraggable.defaultProps = ({
+  isDragDisabled: false,
+  // cannot drag interactive elements by default
+  disableInteractiveElementBlocking: false,
+}: DefaultProps);
+
+export default ConnectedDraggable;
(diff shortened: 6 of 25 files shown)


alexreardon:
Wow! I am off until Wednesday so I’ll take a look then! 👍

TrySound:
- temporary fixed not defined types with eslint `globals`
- moved defaults to `ConnectedDraggable` and `ConnectedDroppable` to fix missing props

.eslintrc
```diff
@@ -12,6 +12,11 @@
     "node": true,
     "jest/globals": true
   },
+  "globals": {
```
SimenB: should be able to use https://github.com/gajus/eslint-plugin-flowtype/blob/d7ccd0ddefbe6589d60f908c90efe453e7604c50/README.md#eslint-plugin-flowtype-rules-define-flow-type instead

package.json
```diff
@@ -74,7 +74,7 @@
     "eslint-plugin-jest": "^21.6.1",
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-react": "^7.5.1",
-    "flow-bin": "0.61.0",
+    "flow-bin": "0.67.1",
```
SimenB: 0.68 is released, FWIW

TrySound: @SimenB I had some problems with it. Gonna try again.

.eslintrc
TrySound: Yep, but seems like it's not supported right now.
https://github.com/gajus/eslint-plugin-flowtype/issues/322

SimenB: That's really odd. I tested adding this snippet to my code base:

```js
somethingGlobal();

const foo: TimeoutID = setTimeout(() => {}, 300);
const bar: IntervalID = setInterval(() => {}, 300);
const baz: AnimationFrameID = requestAnimationFrame(() => {});
```

Runnning eslint on that gives:
```
  20:1   error  'somethingGlobal' is not defined          no-undef
  22:7   error  'foo' is assigned a value but never used  no-unused-vars
  22:12  error  'TimeoutID' is not defined                no-undef
  23:7   error  'bar' is assigned a value but never used  no-unused-vars
  23:12  error  'IntervalID' is not defined               no-undef
  24:7   error  'baz' is assigned a value but never used  no-unused-vars
  24:12  error  'AnimationFrameID' is not defined         no-undef
```

Adding `extends: "plugin:flowtype/recommended"` to the eslint config gives:
```
  20:1  error  'somethingGlobal' is not defined          no-undef
  22:7  error  'foo' is assigned a value but never used  no-unused-vars
  23:7  error  'bar' is assigned a value but never used  no-unused-vars
  24:7  error  'baz' is assigned a value but never used  no-unused-vars
```

No guarantee about the type actually existing, but flow deals with that

src/view/moveable/moveable-types.js
```diff
@@ -12,7 +12,7 @@ export type Props = {|
   speed: Speed,
   destination: Position,
   onMoveEnd: () => void,
-  children: (Style) => Node,
+  children: (Style) => Element<*>,
```
alexreardon: can you explain the reason for this change? We support any React node..

src/view/moveable/moveable.jsx
```diff
@@ -89,9 +89,8 @@ export default class Movable extends Component<Props> {
       // Expecting a flow error
       // React Motion type: children: (interpolatedStyle: PlainStyle) => ReactElement
       // Our type: children: (Position) => (Style) => React.Node
-      // $ExpectError
       <Motion defaultStyle={origin} style={final} onRest={this.onRest}>
-        {(current: Position) =>
+        {(current: { [string]: number }) =>
```
alexreardon: Why can we not use the `Position` type here?

test/unit/view/drag-drop-context.spec.js
```diff
@@ -42,6 +42,10 @@ describe('DragDropContext', () => {
 
     const app = TestUtils.findRenderedComponentWithType(tree, App);
 
+    if (!app) {
```
alexreardon: This will fail silently. In the places in tests where you are checking for null like this can you please throw an error? Something like this:

```js
if(!app) {
 throw new Error('Invalid test setup');
}

File: src/view/droppable/connected-droppable.js [hunk 6]
```diff
@@ -151,13 +153,20 @@ const makeMapStateToProps = () => {
...
   // Droppable gets its own selector
-  makeMapStateToProps,
+  (makeMapStateToProps: any),
   null,
   null,
   { storeKey },
-)(Droppable);
+): any)(Droppable);
 
+connectedDroppable.defaultProps = ({
```
alexreardon: Sad times! We needed this one?

src/view/drag-handle/sensor/create-touch-sensor.js
```diff
@@ -108,7 +108,9 @@ export default ({
   };
 
   const stopPendingDrag = () => {
-    clearTimeout(state.longPressTimerId);
+    if (state.longPressTimerId) {
```
alexreardon: nice

src/state/reducer.js
```diff
@@ -266,14 +266,13 @@ export default (state: State = clean('IDLE'), action: Action): State => {
         return existing;
       }
 
-      // $ExpectError - using spread
```
alexreardon: By casting to `any` are we loosing some type safety? Also, the $ExpectError is very clear about why the following statement fails the check. I think just casting to `any` looses that context. Additionally, hopefully in the future we can remove the ExpectError if flow can understand Object spread a little better. Without further information I think we should move back our spread usage to using ExpectError rather than any.

src/state/get-new-home-client-center.js
```diff
@@ -71,7 +71,6 @@ export default ({
     }
 
     // Otherwise, return the dimension of the empty foreign droppable
-    // $ExpectError - flow does not correctly type this as non optional
```
alexreardon: boya

src/state/reducer.js
alexreardon: This goes for all of these types of changes

package.json
```diff
@@ -74,7 +74,7 @@
     "eslint-plugin-jest": "^21.6.1",
     "eslint-plugin-jsx-a11y": "^6.0.3",
     "eslint-plugin-react": "^7.5.1",
-    "flow-bin": "0.61.0",
+    "flow-bin": "0.68.0",
```
alexreardon: yay!

src/view/droppable/connected-droppable.js
TrySound: This is more honest. This is the component used outside,  so keeping all props required is incorrect.

src/view/moveable/moveable-types.js
TrySound: In this case we inherit react motion types. As a workaround we may add fragment or just fix motion typed. Also `any` may fix this too)

src/view/moveable/moveable.jsx
TrySound: Because the real type of this is more common.

alexreardon: Because we pull out .x and .y i think that using `Position` here is the most clear about what we are dealing with - as the object will always have only these properties

src/view/moveable/moveable-types.js
alexreardon: I'll defer to you for this one

.eslintrc
alexreardon: That that this is needed. This was the thing that made me bail on the upgrade myself. Adding globals seems like the right way to go for now until the eslint-flowtype upgradedhttps://github.com/gajus/eslint-plugin-flowtype/issues/322

src/state/reducer.js
alexreardon: Perhaps we could use [$Exact<>](https://flow.org/en/docs/types/utilities/#exactt-a-classtoc-idtoc-exact-hreftoc-exacta) here

SimenB: `$ExpectError` is not defined by default, leading to unfixable flow errors for consumers. Should be `$FlowFixMe` if you want to ignore it.

I'd recommend removing the custom config here: https://github.com/atlassian/react-beautiful-dnd/blob/4f4c3307a4e0bf066cf8e7dea694d40d8aac9aab/.flowconfig#L10

(`$FlowFixMe` works out of the box)

TrySound: The problem is that flow consider all spreads as inexact. We are loosing the same safety with suppress comment too, but it's more explicit and won't hide different kinds of errors.

src/view/draggable/connected-draggable.js
```diff
@@ -263,12 +264,19 @@ const mapDispatchToProps: DispatchProps = {
...
   // returning a function to ensure each
   // Draggable gets its own selector
-  makeMapStateToProps,
-  mapDispatchToProps,
+  (makeMapStateToProps: any),
+  (mapDispatchToProps: any),
   null,
   { storeKey },
-)(Draggable);
+): any)(Draggable);
```
TrySound: this `any`'s is the result of incomplete react-redux definitions. I'm thinking about solving this problem later.

src/view/moveable/moveable.jsx
TrySound: We inherit this type from react-motion, ignoring this won't help. We can take care of react-motion types later.

src/state/reducer.js
TrySound: $ExpectError should be used only in flow tests.

TrySound:
I think I'm done.

alexreardon:
Great @TrySound ! I will take a look over this morning

alexreardon:
Well done @TrySound!

PR #392, opened 2018-03-19
Add draggable lock

jalkoby:
Hi there. 
Sorry for not creating an issue first (I always read a contribution page just before commit). I think adding ability to lock a draggable element inside his container is nice to have, so here are the code, stories & tests

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -364,3 +364,4 @@ export type Hooks = {|
   onDragEnd: OnDragEndHook,
 |}
 
+export type DraggableLock = 'x' | 'y';
--- a/src/view/draggable/draggable-types.js
+++ b/src/view/draggable/draggable-types.js
@@ -7,6 +7,7 @@ import type {
   Position,
   Direction,
   ZIndex,
+  DraggableLock,
 } from '../../types';
 import {
   lift,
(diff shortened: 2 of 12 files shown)


alexreardon:
Hey @jalkoby,

Thank you for being the change you seek and contributing something that you think is valuable. However, this specific change falls outside the physicality metaphor we are going for.

> #### [Application 5: no drag axis locking](https://github.com/atlassian/react-beautiful-dnd#application-5-no-drag-axis-locking)
> For now, the library does not support drag axis locking (aka drag rails). This is where the user is restricted to only dragging along one axis. The current thinking is this breaks the physical metaphor we are going for and sends a message to the user that they are interacting with a piece of software rather than moving physical objects around. It is possible to ensure that a user can only drop in a single list by using props type and isDropDisabled. You can also do some visual treatment to the list onDragStart to show the user that this is the only place they can interact with.
> -- [Readme](https://github.com/atlassian/react-beautiful-dnd#application-5-no-drag-axis-locking)

For interest sake - you would want to lock the axis on a drag movement rather than while drawing the effect of a drag.

If you are keen to contribute in other ways to this project you can take a look at our [good first issue](https://github.com/atlassian/react-beautiful-dnd/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22) issues!

jalkoby:
Well, you're right - reading is not my choice. Thanks for patient 😅 after all it's good at least pass position on drag

PR #396, opened 2018-03-21
Cancel drag when no visibility #83

### LINKED ISSUE #83: Cancel drag when the page looses focus
It makes sense to cancel a current drag if the page looses focus or visibility (page visiblity api)

grady-lad:
Fixes #83 

### Quick Demo
canceldemo


src/view/drag-handle/util/create-visibility-event.js
```diff
@@ -0,0 +1,42 @@
...
+    return {
+      hidden: 'hidden',
+      name: 'visibilitychange',
+    };
+  }
+  return browserPrefix;
+};
+
+export default (cancel: Function = () => {}) : EventBinding => {
+  if (typeof window !== 'undefined') {
```
alexreardon: When can this function be called when window is not around?

src/view/drag-handle/util/create-visibility-event.js
```diff
@@ -0,0 +1,42 @@
+// @flow
+import type { EventBinding } from './event-types';
+
+const getVisibiltyEvent = () : ?{hidden:string, name:string} => {
+  const prefixes = ['ms', 'webkit', 'moz', 'o'];
+  let browserPrefix;
```
alexreardon: I think we can just use a `find` here (we are already using .find around the place). Also, I think we should check for the default first:

```js
// using memoizeOne so that we do not recompute this function after initial computation
const getVisibilityEvent = memoizeOne() : ?Pair => {
 // non prefixed event is supported
 if('hidden' in document) {
   return {
    hidden: 'hidden',
    name: 'visibilitychange',
  }

  const prefix: ?string = ['ms', 'webkit', 'moz', 'o'].find((prefix: string): boolean => `${prefix}Hidden` in document);

   if(!prefix) {
      return null;
   }

   return {
     hidden: `${prefix}Hidden`,
     name: `${prefix}visibilitychange`,
   }
 }
})

```

test/unit/view/drag-handle.spec.js
```diff
@@ -922,6 +922,23 @@ describe('drag handle', () => {
         expect(event.defaultPrevented).toBe(false);
       });
 
+      it('should cancel when the window is not visibile', () => {
```
alexreardon: can this test please be moved to the shared 'control' section at the bottom so that each sensor type can be tested against this functionality?

src/view/drag-handle/util/create-visibility-event.js
alexreardon:
```js
const defaultPair: Pair = {
    hidden: 'hidden',
    name: 'visibilitychange',
  }

// using memoizeOne so that we do not recompute this function after initial computation
const getVisibilityEvent = memoizeOne() : ?Pair => {
 // non prefixed event is supported
 if('hidden' in document) {
   return defaultPair;

  const prefix: ?string = ['ms', 'webkit', 'moz', 'o'].find((prefix: string): boolean => `${prefix}Hidden` in document);

   // if no prefixed event is supported - simply return the defaultPair. It will not cause any errors to bind to this event - it just won't do anything on visibility change
   if(!prefix) {
      return defaultPair;
   }

   return {
     hidden: `${prefix}Hidden`,
     name: `${prefix}visibilitychange`,
   }
 }
})
```

test/setup.js
```diff
@@ -22,8 +22,10 @@ if (typeof window !== 'undefined') {
 if (typeof document !== 'undefined' && typeof window !== 'undefined') {
   document.documentElement.clientWidth = window.innerWidth;
   document.documentElement.clientHeight = window.innerHeight;
+  // So that we can test window visibility within tests.
+  document.webkitvisibilitychange = () => {};
```
alexreardon: given that we check 'hidden' first we should move to the non prefixed one

test/setup.js
```diff
@@ -22,8 +22,10 @@ if (typeof window !== 'undefined') {
...
+  document.webkitHidden = true;
```
alexreardon: I think it makes sense to have the default has 

```js
document.hidden = false;
```

rather than 'true'.

alexreardon:
Thanks for looking into this @Blasz !

alexreardon:
Thanks for getting into this @grady-lad !!!

src/view/drag-handle/util/create-visibility-event.js
grady-lad: In the integration tests for SSR
Within `test/unit/integration/server-side-rendering.spec.js` the following tests fail
 -  should render identical content when resetting context between renders
 -  should support rendering to static markup
 -  should support rendering to a string

test/setup.js
grady-lad: @alexreardon I am a little stuck about mocking the visibility event for the unit tests.

 The `document` object that is used for the unit tests has a property `hidden` but no `visibilitychange` function. So within `test/setup.js` I added the `visibilitychange` function to the `document` object and within that function tried set `document.hidden = true`.

I presumed that would mock the window being hidden but the function does not get called. 

So I am wondering does anyone have any suggestions how to mock this?

test/unit/view/drag-handle.spec.js
grady-lad: Sure thing, Which section? Is it the `interactive elements` section?

test/setup.js
alexreardon: You can trigger your own events manually. See `user-input-util`

```js
export const dispatchWindowEvent = (eventName: string, options?: Object = {}): Event => {
  const event: Event = document.createEvent('Event');
  event.initEvent(eventName, true, true);

  // override properties on the event itself
  Object.assign(event, options);

  window.dispatchEvent(event);

  return event;
};
```

alexreardon:
This is looking good

alexreardon:
However, it looks like a test is failing in CI. Did you need some help with it?

grady-lad:
Hey @alexreardon I just noticed. 
Yeah that would be great! I am pretty new to flow and haven't seemed to figure out the error.

alexreardon:
I will try to take a look early next week. Happy easter!

alexreardon:
Sorry for the delay on this one. We got totally blindsided by #413. Once things settle a bit I will take another look at this

grady-lad:
@alexreardon no problems, take your time.

alexreardon:
I have updated the PR @grady-lad

grady-lad:
Hey @alexreardon I just seen that changes, thanks for the help 😄. Do you need anything else from my side to close this out?

src/view/drag-handle/sensor/create-keyboard-sensor.js
```diff
@@ -222,6 +223,11 @@ export default ({
       eventName: 'scroll',
       fn: callbacks.onWindowScroll,
     },
+    // Cancel on page visibility change
```
alexreardon: I was able to simply the solution.

alexreardon: We do not need to know what the visibility of the page is when the even fires - we simply need to know if the event fires. If it fires while a drag is occurring we can kill the drag

alexreardon:
Thanks for your patience @grady-lad! Well done on your first contribution

PR #398, opened 2018-03-22
add type to OwnProps in draggable

MateuszLeo:
add type to OwnProps in draggable


alexreardon:
A draggable pulls it's type from it's parent droppable. Right now we do not support a draggable being a child of a droppable that is of a different type

PR #401, opened 2018-03-22
correctly blocking SVG elements within interactive elements

### LINKED ISSUE #399: Nested interactive elements : Button with svg
## Bug
After #359 issue and the 6.0.0 update, it works great with only "text" child.

But, a button like this isn't blocked :
```
<button>
    <div>
        <svg>
            <path>
        </svg>
    </div>
</button>
```

### Expected behavior
Same as #359 

> Dragging should be blocked no matter what button contains. Useful for svg icons for example.

### Actual behavior
I can drag an element when I click on a button with a `svg` inside a `div`.

### Steps to reproduce
Make a button with an `svg` tag inside.
```
<button>
    <div>
        <svg>
            <path>
        </svg>
    </div>
</button>
```

### Browser version
Chrome 65.0.3325.181 64 bits

alexreardon:
Fixes #399


src/debug/timings.js
```diff
@@ -10,11 +10,6 @@ const flag: string = '__react-beautiful-dnd-debug-timings-hook__';
 
 const isTimingsEnabled = (): boolean => Boolean(window[flag]);
 
-if (process.env.NODE_ENV === 'development') {
```
alexreardon: removing this also for now :)

src/view/drag-handle/util/should-allow-dragging-from-target.js
```diff
@@ -59,7 +59,7 @@ export default (event: Event, props: Props): boolean => {
   const { target, currentTarget } = event;
 
   // Technically target and currentTarget are EventTarget's and do not have to be elements
-  if (!(target instanceof HTMLElement) || !(currentTarget instanceof HTMLElement)) {
+  if (!(target instanceof Element) || !(currentTarget instanceof Element)) {
```
alexreardon: this was the fix

test/unit/view/drag-handle.spec.js
```diff
@@ -2919,7 +2925,69 @@ describe('drag handle', () => {
...
+                control.lift(wrapper, options);
+                control.drop(wrapper);
+
+                expect(callbacksCalled(callbacks)({
+                  onLift: 0,
+                })).toBe(true);
+              });
+            });
+
+            it('should start a drag on a Element with an interactive parent if asked to by user', () => {
```
Blasz: I think this test name should be swapped with the one below.

alexreardon: Cheers

jquense:
Can someone speak to why this change was made? We had all our drag handles stop working because they use svg icons.

alexreardon:
Hi @jquense, we currently block dragging of interactive elements by default. Elements such as `button`, `input` and so on. If we do not do this then those elements are useless as interactive elements as the user can never interact with them.

This change should not prevent you from having an SVG in a drag handle! It will only block if you drag handle is a `button` or some other interactive element. 

You can opt out of this behaviour by using the `disableInteractiveElementBlocking` prop on a `Draggable`. 

[More information](https://github.com/atlassian/react-beautiful-dnd#interactive-child-elements-within-a-draggable)

PR #402, opened 2018-03-23
updating deps and build scripts

alexreardon:
Bumping a few internal deps
Also putting our generated flow file into the `esm` folder


src/view/drag-handle/sensor/create-keyboard-sensor.js
```diff
@@ -18,8 +18,8 @@ type State = {|
 |}
 
 type ExecuteBasedOnDirection = {|
-  vertical: () => void,
-  horizontal: () => void,
+  vertical: Function,
+  horizontal: Function,
```
Blasz: Flow recommends [not using Function](https://flow.org/en/docs/types/functions/#toc-function-type), did you have issues with `() => void` or `() => mixed`?

alexreardon: Some functions return a `AnimationFrameID`. I can move to () => mixed but i think just simply saying it is a `Function` is sufficient for our purposes here

PR #405, opened 2018-03-24
Combine rollup configs with multiconfig and add size snapshot

TrySound:
- combined both umd/min configs with rollup multiconfig
- changed umd bundle to development env like react does
- added my [size-snapshot](https://github.com/TrySound/rollup-plugin-size-snapshot) plugin to simplify size tracking and improving in some cases. Later we can probably bundle also cjs and esm to track their treeshakability.


alexreardon:
I am trying to think of what the development flow would be with this new plugin. Can you please elaborate on this? Also, how would we go about rebase lining?

alexreardon:
Okay, so you would need to run the command to generate the size snapshot. Would it be good to have a seperate test script which simply checks to see if the bundle size has grown? That way we could run it as a build check

TrySound:
@alexreardon Done.

package.json
```diff
@@ -25,15 +25,15 @@
     "/src"
   ],
   "scripts": {
-    "test": "jest --config ./jest.config.js",
+    "test": "cross-env SNAPSHOT=check yarn build:dist && jest --config ./jest.config.js",
```
alexreardon: is there any way we can run the test without generating the dist itself? hmm...

TrySound: AFAIK there's no way to prevent rollup writing currently. Is this a problem?

TrySound: The idea of the plugin was integration in bundler without additional configuration.

alexreardon:
👌

TrySound:
@alexreardon Can we consider using bundles for cjs and esm before major 7 release. I'd like to open PR after this one get merged.

package.json
alexreardon: Can this create a scenario where a merge to master can break? Thinking out loud.. i suppose not if it it updating the snapshot too

alexreardon: Can we run the check as the second step of the command? (After jest)

TrySound: @alexreardon It can if PR is not based on actual master. But I think it's not so critical.

TrySound:
Done

alexreardon:
Well done @TrySound! 

We'll see how the snapshot watching goes. If it causes too much friction we *might* remove it at some point. However, for now I am keen to have more visibility around our bundle size changes

PR #407, opened 2018-03-28
Add puppeeter simple test

RajaBellebon:
* Add a single drag test for keyboard using single list


.travis.yml
```diff
@@ -24,5 +24,7 @@ env:
 # Required for lock file generation
 # https://github.com/greenkeeperio/greenkeeper-lockfile
 before_install: yarn global add greenkeeper-lockfile@1
-before_script: greenkeeper-lockfile-update
+before_script: 
+  - greenkeeper-lockfile-update
+  - yarn storybook & sleep 7
```
alexreardon: `sleep 7`? any way to make this a bit more deterministic?

test/utils/css.js
```diff
@@ -0,0 +1,7 @@
+// @flow
```
alexreardon: for now let's move these variables into our only test file

test/integration/single-list.spec.js
```diff
@@ -0,0 +1,60 @@
+// @flow
```
alexreardon: Given we are starting a browser in this test, I think it best to add the node pragma to the top of this file:

```js
/**
* @jest-environment node
*/
// @flow
```

That way we can be sure we are not touching jsdom

TrySound:
This is quite expensive setup which creates puppeteer instance in every `spec`. I recommend to use [jest-environment-puppeteer](https://github.com/smooth-code/jest-puppeteer/tree/master/packages/jest-environment-puppeteer)

TrySound:
Here's examples how I test with puppeteer
https://github.com/renatorib/react-powerplug/blob/master/tests/components/FocusManager.test.js
https://github.com/renatorib/react-powerplug/blob/master/package.json#L36-L37

.travis.yml
RajaBellebon: It is a bit difficult as you don't know how long will it take to be built. I found this workaround when I used cypress 

I can try another approach `-- --silent &`

test/utils/css.js
RajaBellebon: Ok - no problem, from experience - I know that it is always better clubbing all selectors in one file but as we have only one that's fine for now

.travis.yml
lukebatchelor: Previously we've used something to keep trying to ping the server until it responds

RajaBellebon: Do you still have the command?

RajaBellebon: hmm - even in Travis documentation - they still use sleep:
https://docs.travis-ci.com/user/gui-and-headless-browsers/

test/integration/single-list.spec.js
RajaBellebon: @TrySound - this was my original approach but after discussing with @alexreardon  and as we have only 1 spec file for now when we will have more than 1 spec it makes more sense to have  jest-environment-puppeteer.

alexreardon:
Perhaps we could add jest-environment-puppetter as a follow up PR?

.travis.yml
```diff
@@ -24,5 +24,8 @@ env:
 # Required for lock file generation
 # https://github.com/greenkeeperio/greenkeeper-lockfile
 before_install: yarn global add greenkeeper-lockfile@1
-before_script: greenkeeper-lockfile-update
-after_script: greenkeeper-lockfile-upload
\ No newline at end of file
+before_script: 
+  - greenkeeper-lockfile-update
+  - yarn global add wait-port@0.2.2
+  - yarn storybook & wait-port -t 10000 localhost:9002
```
alexreardon: So good

emanuellarini:
I could make a drag with this code. 

Actually, my project is a drag n drop checkers game.

`
describe('The Game Movements', () => {
  test('player 1 can move its disc to the upper right square towards opponent', async () => {
    const browser = await puppeteer.launch({
      headless: false,
      slowMo: 100,
      args: [
        `--window-size=${ 1440 },${ 900 }`
      ]
    })
    const page = await browser.newPage()
    await page.setViewport({width: 1440, height: 900, deviceScaleFactor: 1})
    await page.goto('http://localhost:3000')
    await page.waitFor(900)

    const dropSquare = await page.waitForSelector('[data-testid="board-square-4-3"]');
    const player1Disc = await page.waitForSelector('[data-testid="board-square-5-2"]');
    const dropSquareBox = await dropSquare.boundingBox();
    const player1DiscBox = await player1Disc.boundingBox();

    await page.mouse.move(player1DiscBox.x + player1DiscBox.width/2, player1DiscBox.y + player1DiscBox.height/2);
    await page.mouse.down();
    await page.mouse.move(dropSquareBox.x + dropSquareBox.width/2, dropSquareBox.y + dropSquareBox.height/2, {steps: 6});
    await page.mouse.up();

    // expectations... :)

    await browser.close();
  })
})
`

Considerations:
1. The first waitFor is for my animations, you shouldn't need that
2. I had to use slowMo to slow down Puppeteer operations
3. Each squares of my board in 1440x900 viewport has a size of 60x60.
4. I calculated the disc positions X and Y, and desired drop position X and Y.
5. Mouse moves to draggable disc position, mouse button is pressed down, mouse moves to the desired droppable square and mouse button is pressed up
6. Profit

PR #416, opened 2018-04-03
iOS 11.3 fix

### LINKED ISSUE #413: ios 11.3 not supported
## Bug or feature request?

Bug

### Expected behavior

Drag and drop works beautifully as usual.

### Actual behavior

With the upgrade to ios 11.3, drag and drop is not beautiful.  When attempting to select and drag, the page now scrolls making it almost impossible to move an item.

### Steps to reproduce

Upgrade to ios 11.3 and attempt to drag and drop.  

### Browser version

Tried Chrome 65.0.3325.152 and whatever Safari version ios 11.3 ships with (Safari 11).

### Demo

Tested on our app and also`react-beautiful-dnd` mobile examples provided (show in video below):

react_beautiful_dnd_ios_11 3

Thanks for your work on this excellent library!

alexreardon:
Fixes #413 

This is a hack to get things working again. I cannot seem to find the exact root cause. There is a bug that I have noticed and raised with webkit. If that is fixed then I can remove the hack. However, my find does not explain why it worked on iOS 11.2. I am trying really hard to find an answer

### DIFF
--- a/src/view/drag-handle/sensor/create-touch-sensor.js
+++ b/src/view/drag-handle/sensor/create-touch-sensor.js
@@ -23,11 +23,74 @@ type TouchWithForce = Touch & {
   force: number
 }
 
+type SafariHack = {|
+  start: () => void,
+  stop: () => void,
+|}
+
 export const timeForLongPress: number = 150;
 export const forcePressThreshold: number = 0.15;
 const touchStartMarshal: EventMarshal = createEventMarshal();
-
 const noop = (): void => { };
+
+// Safari does not allow event.preventDefault() in dynamically added handlers
+// So we add an always listening event handler to get around this :(
+// webkit bug: https://bugs.webkit.org/show_bug.cgi?id=184250
+const safariHack: SafariHack = (() => {
+  const shouldBlock: boolean = (() => {
+    // Do nothing when server side rendering
+    if (typeof window === 'undefined') {
+      return false;
+    }
+
+    const isUsingSafari11: RegExp = /AppleWebKit.*Version\/11/g;
+
+    // Not using Safari 11
+    if (!isUsingSafari11.test(window.navigator.userAgent)) {
+      return false;
+    }
+
+    // Using Safari 11 with no touch support - no point adding the touch listeners
+    if (!('ontouchstart' in window)) {
+      return false;
+    }
+
+    return true;
+  })();
+
+  if (!shouldBlock || typeof window === 'undefined') {
+    return { start: noop, stop: noop };
+  }
+
+  let isBlocking: boolean = false;
+  const start = () => {
+    isBlocking = true;
+  };
+  const stop = () => {
+    isBlocking = false;
+  };
+
+  // Adding a persistent event handler
+  window.addEventListener('touchmove', (event: TouchEvent) => {
+    if (!isBlocking) {
+      return;
+    }
+
+    // Our normal event handler should have done this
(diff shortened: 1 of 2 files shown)


src/view/drag-handle/sensor/create-touch-sensor.js
```diff
@@ -135,10 +135,52 @@ export default ({
...
+  // Safari does not allow event.preventDefault() in dynamically added handlers
+  // So we add an always listening event handler to get around this :(
+  // webkit bug: https://bugs.webkit.org/show_bug.cgi?id=184250
+  (() => {
+    // Do nothing when server side rendering
+    if (typeof window === 'undefined') {
+      return;
+    }
+
+    const isUsingSafari11: RegExp = /AppleWebKit.*Version\/11/g;
```
alexreardon: You know it is a bad day when you need to do some browser sniffing 😢

src/view/drag-handle/sensor/create-touch-sensor.js
```diff
@@ -135,10 +135,52 @@ export default ({
...
+      return;
+    }
+
+    // Using Safari 11 with no touch support - no point adding the touch listeners
+    if (!('ontouchstart' in window)) {
+      return;
+    }
+
+    // Adding a persistent event handler
+    window.addEventListener('touchmove', (event: TouchEvent) => {
```
Blasz: Is there any part of the lifecycle where we would want to remove the event listener? E.g. on unmount?

src/view/drag-handle/sensor/create-touch-sensor.js
```diff
@@ -23,12 +23,75 @@ type TouchWithForce = Touch & {
...
+
+    // Using Safari 11 with no touch support - no point adding the touch listeners
+    if (!('ontouchstart' in window)) {
+      return false;
+    }
+
+    return true;
+  })();
+
+  if (!shouldBlock || typeof window === 'undefined') {
```
Blasz: Nit: The `typeof window` part of the conditional won't get hit since it's being checked within `shouldBlock`.

src/view/drag-handle/sensor/create-touch-sensor.js
```diff
@@ -23,12 +23,77 @@ type TouchWithForce = Touch & {
...
+// webkit bug: https://bugs.webkit.org/show_bug.cgi?id=184250
+const webkitHack: WebkitHack = (() => {
+  // Do nothing when server side rendering
+  if (typeof window === 'undefined') {
+    return { preventTouchMove: noop, releaseTouchMove: noop };
+  }
+
+  const shouldBlock: boolean = (() => {
+    // All browsers on iPhone or iPad
+    const target: RegExp = /OS\s11_\d\slike\sMac\sOS\sX/g;
```
alexreardon: As discussed with @seancurtis, I will remove this check. This will match against iphones and ipads - but not other webkit based browsers. Given that the cost of this is a single event listener I am okay to add it redundantly to other browsers which do not strictly need it in order to avoid other webkit browsers slipping through the cracks

rohan-deshpande:
The frustration in your comments is understandable! I’m currently considering migrating from Dragula to this package because I cannot get this to work with it. I’d much prefer to not do this though. I’m wondering if you might provide a vanilla js explanation for your thinking behind this implementation, at a glance, I get it but I’ve not seen the `capture: false` in anyone else’s proposed solution.

Issue #192, opened 2017-11-22
Move to portal solution for Draggables


onemenny:
## Bug or feature request?
Feature request

Well, it is lame, as mentioned in the ReadMe. I had to deal with transitionend event, plus add many patches to get it semi working. I do believe that transitions have become a standard with animations and more. Having a transitioned panel as a parent of beautiful-dnd is not far fetched, as I had to deal with it. Append to body option is something many 3rd party component adds to give a 360 degree compliance 

I for one, would certainly like to see this feature added

Any way, thank you for the hard work.

alexreardon:
For context:

> Warning: position: fixed
> 
> `react-beautiful-dnd` uses `position: fixed` to position the dragging element. This is quite robust and allows for you to have `position: relative | absolute | fixed` parents. However, unfortunately `position:fixed` is [impacted by `transform`](http://meyerweb.com/eric/thoughts/2011/09/12/un-fixing-fixed-elements-with-css-transforms/) (such as `transform: rotate(10deg);`). This means that if you have a `transform: *` on one of the parents of a `Draggable` then the positioning logic will be incorrect while dragging. Lame! For most consumers this will not be an issue. We may look into creating a portal solution where we attach the dragging element to the body rather than leave it in place. However, leaving it in place is a really nice experience for everyone. For now we will leave it as is, but feel free to raise an issue if you this is important to you.

Now that react has a [supported portal feature](https://reactjs.org/docs/portals.html) we could look into doing it. I am keen to avoid any regressions on the consumption experience which is currently really nice. We will probably not look at this for a while as we have a large number of higher priority features we want to ship. In the mean time you might find this issue helpful: #128 specifically the part about overriding top/left values rather than using the transform.

onemenny:
Thank, I was aware of your constraints. I will look into it.

kasperpihl:
I managed to get this to work with Portal from react-dom without changing styles etc.

I added a dom element into my html

```
<div id="draggable"></div>
```

Added the following styles
```
position: absolute;
pointer-events: none;
height: 100%;
width: 100%;
```

And then added this to the Draggable
```
import React, { PureComponent } from 'react';
import { createPortal } from 'react-dom';

const _dragEl = document.getElementById('draggable');

class DraggableGoal extends PureComponent {
  optionalPortal(styles, element) {
    if(styles.position === 'fixed') {
      return createPortal(
        element,
        _dragEl,
      );
    }
    return element;
  }
  render() {
[...]

PR #421, opened 2018-04-04
support transformed draggables without portals

mattkrick:
Per #192 it looks like portals are not the way to go.
I agree. I usually love portals, but when performance is critical, they aren't great.
I have a hard requirement on using a transform.
We _could_ do this without extending the API, but that would be more expensive because we'd have to climb the DOM tree looking for a `transform`. In the spirit of performance-first, I opted to add a single extra `provided` function. For those that don't use transforms, no change is required. For those that do, they add a single argument to a div.

How it works:
react-beautiful-dnd uses `position: fixed`, which is great, except for 1 caveat. Per MDN docs:

> When an ancestor has the transform, perspective, or filter property set to something other than none, that ancestor is used as the container instead of the viewport

If the developer provides a `transformRef`, then that tells the `Draggable` that an ancestor is being transformed, and therefore the `left, top` should be relative to the transformed node, _not_ the viewport. It adjusts the `left, top` accordingly.


abcNocturn:
Would really like to see this realeased as soon as possible ...
Got a transformed Element Dialog from MaterialUI which is unusable with react-dnd ....

alexreardon:
We are about to release 7.0 which changes things. Feel free to create another PR to readd this change to 7.0. 

Perhaps a different way to go is to implement this as a (/docs/patterns). I suspect this can be implemented as a consuming rather than needing to add it to api.

Our current recommendation is to use `React.Portal` if you cannot get around having a parent with a clashing style.

I am closing this PR as it is not in the state that we want to support. As mentioned, feel free to raise another

PR #429, opened 2018-04-09
Provide cjs and esm bundles with information about treeshakability

TrySound:
Ref #405 
Closes https://github.com/atlassian/react-beautiful-dnd/pull/428

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,13 +15,11 @@
   "bugs": {
     "url": "https://github.com/atlassian/react-beautiful-dnd/issues"
   },
-  "main": "lib/index.js",
-  "module": "esm/index.js",
+  "main": "dist/react-beautiful-dnd.cjs.js",
+  "module": "dist/react-beautiful-dnd.esm.js",
   "sideEffects": false,
   "files": [
-    "/lib",
     "/dist",
-    "/esm",
     "/src"
   ],
   "scripts": {
@@ -29,12 +27,10 @@
     "validate": "yarn run lint && yarn run typecheck",
     "lint": "yarn eslint .",
     "typecheck": "flow check",
-    "build": "yarn run build:clean && yarn run build:lib && yarn run build:esm && yarn run build:flow && yarn run build:dist",
-    "build:clean": "rimraf lib && rimraf esm && rimraf dist",
-    "build:lib": "cross-env NODE_ENV=cjs babel src -d lib",
-    "build:esm": "babel src -d esm",
+    "build": "yarn run build:clean && yarn run build:dist && yarn run build:flow",
+    "build:clean": "rimraf dist",
     "build:dist": "rollup -c",
-    "build:flow": "echo \"// @flow\n\nexport * from '../src';\" | tee lib/index.js.flow esm/index.js.flow",
+    "build:flow": "echo \"// @flow\n\nexport * from '../src';\" > dist/react-beautiful-dnd.cjs.js.flow",
     "storybook": "start-storybook -p 9002",
     "build-storybook": "build-storybook -c .storybook -o site",
     "prepublish": "yarn run build && cd website && yarn",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -15,26 +15,22 @@
   "bugs": {
     "url": "https://github.com/atlassian/react-beautiful-dnd/issues"
   },
-  "main": "lib/index.js",
-  "module": "esm/index.js",
+  "main": "dist/react-beautiful-dnd.cjs.js",
```
alexreardon: is it common to do this rather than use a seperate folder?

alexreardon: Dumb question, but cjs = CommonJS? 😊

package.json
```diff
@@ -15,26 +15,22 @@
...
     "typecheck": "flow check",
-    "build": "yarn run build:clean && yarn run build:lib && yarn run build:esm && yarn run build:flow && yarn run build:dist",
-    "build:clean": "rimraf lib && rimraf esm && rimraf dist",
-    "build:lib": "cross-env NODE_ENV=cjs babel src -d lib",
-    "build:esm": "babel src -d esm",
+    "build": "yarn run build:clean && yarn run build:dist && yarn run build:flow",
+    "build:clean": "rimraf dist",
     "build:dist": "rollup -c",
-    "build:flow": "echo \"// @flow\n\nexport * from '../src';\" | tee lib/index.js.flow esm/index.js.flow",
+    "build:flow": "echo \"// @flow\n\nexport * from '../src';\" > dist/react-beautiful-dnd.cjs.js.flow",
```
alexreardon: Is flow okay with this? Do we want it also for the .esm.js file too?

rollup.config.js
```diff
@@ -47,6 +51,36 @@ const getUMDConfig = ({ env, file }) => {
 };
 
 export default [
-  getUMDConfig({ env: 'development', file: 'dist/react-beautiful-dnd.js' }),
+  getUMDConfig({ env: 'development', file: 'dist/react-beautiful-dnd.umd.js' }),
+
   getUMDConfig({ env: 'production', file: 'dist/react-beautiful-dnd.min.js' }),
```
alexreardon: .umd.min.js?

package.json
TrySound: I think it is not necessary to create a folder for only one file in it. We have five file in the end which look very nice in one folder.

Yes, esm and cjs are the most popular shorthands.

package.json
TrySound: I don't think it's necessary. Users usually do not specify esm entry directly, but rely on bundler.

rollup.config.js
TrySound: This will break pattern :(

```
react-beautiful-dnd.cjs.js
react-beautiful-dnd.cjs.js.flow
react-beautiful-dnd.esm.js
react-beautiful-dnd.min.js
react-beautiful-dnd.umd.js
```

I just would like to see this pattern widespread, so `min` in most cases could mean `umd` format. I already used it in a set of modules.

alexreardon:
Can you please elaborate on:

> Preventing user from importing internals.

TrySound:
Users may use some internals with `import something from 'react-beautifull-dnd/dist/something.js'`. It's usually a private code, which should not be considered as public API. Bundle allows to isolate such private code in single module scope, so users won't have an access to it and won't be broken when file structure is changed.

alexreardon:
Perfect. Thanks

rollup.config.js
```diff
@@ -8,10 +8,14 @@ import replace from 'rollup-plugin-replace';
 import strip from 'rollup-plugin-strip';
 import { sizeSnapshot } from 'rollup-plugin-size-snapshot';
 
+const pkg = require('./package.json');
+
 const input = './src/index.js';
 
 const extensions = ['.js', '.jsx'];
 
+const isExternal = id => !id.startsWith('.') && !id.startsWith('/');
```
alexreardon: can you please add a comment above this line in the code explaining its purpose? Other than that I think I am good to go!

rollup.config.js
```diff
@@ -47,6 +51,36 @@ const getUMDConfig = ({ env, file }) => {
...
+    input,
+    output: {
+      file: pkg.module,
+      format: 'es',
+    },
+    external: isExternal,
+    plugins: [
+      resolve({ extensions }),
+      babel(getBabelOptions()),
+      strip({ debugger: true }),
```
alexreardon: I am keen to keep console.* statements for the non-minified builds. Can the `strip` function be updated so that only the minified build strips console.*? https://github.com/rollup/rollup-plugin-strip#usage

TrySound: From non minified umd too? It seems console was always stripped from both umd bundles

TrySound:
Done

alexreardon:
Well done @TrySound. Thanks for your efforts in getting this in!

PR #438, opened 2018-04-12
Fix typo in README

christianhg:
I'm pretty sure this is supposed to be "true" instead of "false". However, I wasn't entirely sure if it should read "true" or "`true`". I can amend the commit if I made the wrong decision.


alexreardon:
Thanks for this @christianhg !!

Issue #437, opened 2018-04-12
Move from react-motion to react-spring?


alexreardon:
It would be worth taking a look at [react-spring](https://github.com/drcmda/react-spring) to see if we can get any performance wins while maintaining our feeling of physicality

souporserious:
I'd love to hear your thoughts on [Popmotion](https://popmotion.io/). They have a [`spring` implementation](https://popmotion.io/api/spring/) that can be imported directly which I think could reduce the bundle from where it is now. I'd be willing to start a PR if you are interested.

TrySound:
I'm popmotion doesn't have esm bundle and esm path imports. I could work on it if they are opened for changes.

souporserious:
I think Matt would be open to that! cc: @InventingWithMonster

alexreardon:
Worth investigating for sure!!!

alexreardon:
Before we go down the esm path I think it worth seeing how using the
libraries feel as a replacement for react-motion for the dragging item

alexreardon:
Criteria:

- We want to maintain the same physics characteristics as the current react-motion usage for the dragging item. It feels amazing when dropping
- Ideally we want to use something as light as possible: light on the CPU and smaller KB

TrySound:
KB is not absolute criteria. Find something and I will investigate how small it can be.

InventingWithMonster:
Is there good reading on what a ESM bundles are, how they're made and why I should implement them? I'm definitely open to anything that makes it easier for people to use Popmotion, it's just at the moment every animation is importable individually and I assumed that was enough for people to consume the library in a lightweight manner.

You might find some joy with [Pose](https://popmotion.io/pose) which handles all the FLIP stuff too and works with the DOM directly. This would be easier to bundle but if you want to tweak the spring (and you probably do) you need to import it from Popmotion hence why I think bundling may be problematic.

TrySound:
@InventingWithMonster If everything in your code is a function or a class without static properties then everything is treeshakable. On the other side transpiled to esm files (with `modules: false` option) can also significantly decrease the size of the final bundle because this code won't be wrapped with commonjs runtime.

InventingWithMonster:
@TrySound Thanks for the insight! It looks easy enough to add transpilation out into into an `es` (or a standardised name?) directory with TypeScript `module="none"`. I'll have time over the weekend.

Is there an easy way in package.json to flag the existence of this directory or is it something users have to manually point to?

[...]

PR #442, opened 2018-04-16
replace react-motion with react-spring

drcmda:
This PR removes react-motion and replaces it with react-spring (see: https://github.com/atlassian/react-beautiful-dnd/issues/437).

Since the api is almost the same it was pretty straight forward. I removed react-motion specific work arounds as well.

In a later stage we can think about how to enable native rendering.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,21 +1,21 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 399077,
-    "minified": 155345,
-    "gzipped": 46468
+    "bundled": 346612,
+    "minified": 136359,
+    "gzipped": 39394
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 360429,
-    "minified": 138877,
-    "gzipped": 41518
+    "bundled": 308115,
+    "minified": 120270,
+    "gzipped": 34685
   },
   "dist/react-beautiful-dnd.esm.js": {
-    "bundled": 176172,
-    "minified": 88689,
-    "gzipped": 22490,
+    "bundled": 176294,
+    "minified": 88740,
+    "gzipped": 22529,
     "treeshaked": {
-      "rollup": 77412,
-      "webpack": 78994
+      "rollup": 77453,
+      "webpack": 79052
     }
   }
 }
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,21 +1,21 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 372617,
-    "minified": 142790,
-    "gzipped": 40319
+    "bundled": 388896,
+    "minified": 154817,
+    "gzipped": 46416
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 333972,
-    "minified": 126619,
-    "gzipped": 35549
+    "bundled": 350133,
+    "minified": 138467,
+    "gzipped": 41546
   },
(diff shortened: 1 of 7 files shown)


drcmda:
Also, `translate` should probably be translate3d so we get the gpu to kick in? Or was that a deliberate choice?

TrySound:
Test will ask you to update size snapshot. You can do this by running `yarn build` script.

drcmda:
As for the tests flagged red, now i don't know if these test against the react-motion workaround i saw in the code or if it's something else. I'd be glad to receive help here. 

@TrySound I suggest once you're happy with your optimizations we could make a build, so we get the benefits of a small, tree-shaken build.

drcmda:
Fixed a bunch of complaints. Scratching my head over the remaining few, where it expects it to sit at 0/0 but it's at 75/75 - was that the react-motion bug?

Also added precision to decrease the time it takes for the element to settle. Can be tweaked some more.

.size-snapshot.json
```diff
@@ -1,21 +1,21 @@
...
-    "gzipped": 35549
+    "bundled": 360429,
+    "minified": 138877,
+    "gzipped": 41518
   },
   "dist/react-beautiful-dnd.esm.js": {
-    "bundled": 176774,
-    "minified": 88946,
-    "gzipped": 22589,
+    "bundled": 176172,
```
alexreardon: moving to react-spring would actually result in a minor drop of kb's 👍 

not for the UMD build but i am not too worried about that one

TrySound: I already achieved better results in master. UMD build is bigger because of babel runtime helpers which are not reused.

drcmda:
Some interesting development in here: https://github.com/drcmda/react-spring/pull/48

Still i have no idea how to delegate `native` to the end-user, but if there's a solution one day, then with that pr not even mousemove will cause rendering, the Spring would basically bind a `Value` object to a view, and steers it outside of the render loop and mouse/touch events.

alexreardon:
I am keen to see this one progress. It is looking really promising

alexreardon:
Given we can get rid of a lot of hacks without any KB hit it seems like a good move

TrySound:
@drcmda Update dependencies please. Your size results are not relevant.

drcmda:
@TrySound @alexreardon Found a way to isolate the spring, even remove it from interpolation. The result is ~3kb, basically it's just the math for the spring now. No element names, color interpolation, etc. NumericalSpring is now part of addons, good for cases like this where we only need to animate raw numerical data. It behaves like Spring otherwise with support for configs, native and so on.

Theoretically it can be reduced even further, if onRest/tracking/onFrame isn't needed.

alexreardon:
Whoa

TrySound:
@drcmda Can you rebase and update dependencies, so we could get real size difference.

alexreardon:
I am having a look at how the library feels now

alexreardon:
I had an initial play with this and things look positive. I actually noticed some minor performance regressions, but i am not sure what that is about. 

This feels like a good move forward as it:

- Removes a lot of hacky react-motion work arounds
- Has an expressive component api
- Seems to work quite nicely with the exisiting setup
- moves to a more maintained project

Things we still need to do:

- I noticed some bugs when dragging cards on examples that had multiple columns
- Further investigate performance 
- Update tests
- Remove all react-motion hacks (i think there is still one in the action creators)

I am keen to see somebody take this up. Unfortunately I am working on a few other things (get excited) so I do not have capacity right now to own this

PR #443, opened 2018-04-16
replace react-motion with wobble

souporserious:
I had about 20 minutes to see if I could get this working so it's not perfect and the spring animation isn't working properly, but I can come back to it if the smaller size difference seems worth it.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,21 +1,21 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 372617,
-    "minified": 142790,
-    "gzipped": 40319
+    "bundled": 321260,
+    "minified": 128627,
+    "gzipped": 37326
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 333972,
-    "minified": 126619,
-    "gzipped": 35549
+    "bundled": 282830,
+    "minified": 112597,
+    "gzipped": 32657
   },
   "dist/react-beautiful-dnd.esm.js": {
-    "bundled": 176774,
-    "minified": 88946,
-    "gzipped": 22589,
+    "bundled": 176607,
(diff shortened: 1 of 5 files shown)


.size-snapshot.json
```diff
@@ -1,21 +1,21 @@
...
-    "bundled": 176774,
-    "minified": 88946,
-    "gzipped": 22589,
+    "bundled": 176607,
+    "minified": 88967,
+    "gzipped": 22548,
     "treeshaked": {
-      "rollup": 77588,
-      "webpack": 79186
+      "rollup": 77702,
```
alexreardon: Interesting that the tree shaked size actually went up. There must be a lot of dead code elimination in react-motion

.size-snapshot.json
```diff
@@ -1,21 +1,21 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 372617,
-    "minified": 142790,
-    "gzipped": 40319
+    "bundled": 321260,
```
alexreardon: nice drop here

.size-snapshot.json
```diff
@@ -1,21 +1,21 @@
...
+    "minified": 112597,
+    "gzipped": 32657
   },
   "dist/react-beautiful-dnd.esm.js": {
-    "bundled": 176774,
-    "minified": 88946,
-    "gzipped": 22589,
+    "bundled": 176607,
+    "minified": 88967,
+    "gzipped": 22548,
```
alexreardon: only a tiny kb improvement at the end of the day on react-motion

souporserious: Yeah, it seemed very minute so I wasn't sure if you'd want to move forward with it or not. If it seems worth it, please let me know and I can fix the PR up and get tests passing.

PR #446, opened 2018-04-16
Upgrading deps

alexreardon:


.size-snapshot.json
```diff
@@ -1,21 +1,21 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 372617,
```
alexreardon: Main code change i can see is our moving from `throw new Error` to `invariant` in an attempt to reduce kb's in tree shaking. However, it looks like our umd bundle went up. Any thoughts @TrySound ?

src/types.js
```diff
@@ -364,15 +364,15 @@ export type DropState = {|
   result: ?DropResult,
 |}
 
-export type State = {
+export type State = {|
```
alexreardon: now an exact type

.size-snapshot.json
TrySound: Don't know to be honest.

@lukastaegert what was changed in the latest versions?

.size-snapshot.json
```diff
@@ -1,21 +1,21 @@
...
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 334038,
-    "minified": 126662,
-    "gzipped": 35556
+    "bundled": 360109,
+    "minified": 132835,
+    "gzipped": 37619
   },
   "dist/react-beautiful-dnd.esm.js": {
```
alexreardon: our esm builds are looking better for moving to invariant rather than throwing, but otherwise we took a huge hit!

.size-snapshot.json
alexreardon: I updated the snapshot on master so we are sure we are looking at the correct thing

alexreardon: I had a look at the build and it is not creating multiple copies of invariant..

alexreardon: I think it might be a transitive dependency change..

alexreardon: Okay, i reversed the invariant usages changes and i can confirm that the size changes did not come from there

yarn.lock
```diff
@@ -1637,9 +1655,9 @@ babylon@7.0.0-beta.31:
   version "7.0.0-beta.31"
   resolved "https://registry.yarnpkg.com/babylon/-/babylon-7.0.0-beta.31.tgz#7ec10f81e0e456fd0f855ad60fa30c2ac454283f"
 
-babylon@7.0.0-beta.42, babylon@^7.0.0-beta.40:
-  version "7.0.0-beta.42"
-  resolved "https://registry.yarnpkg.com/babylon/-/babylon-7.0.0-beta.42.tgz#67cfabcd4f3ec82999d29031ccdea89d0ba99657"
+babylon@7.0.0-beta.44:
```
alexreardon: This might also be the cause

yarn.lock
```diff
@@ -7298,9 +7326,12 @@ rollup-pluginutils@^2.0.1:
     estree-walker "^0.3.0"
     micromatch "^2.3.11"
 
-rollup@^0.56.2:
-  version "0.56.5"
-  resolved "https://registry.yarnpkg.com/rollup/-/rollup-0.56.5.tgz#40fe3cf0cd1659d469baad11f4d5b6336c14ce84"
+rollup@^0.58.0:
```
alexreardon: this could have caused it

alexreardon: nah i just checked

lukebatchelor:
So, looking at the diff of the two bundles (before and after this PR) I get an interesting result.

Immediately I see the new bundle has invariant bundled at the top and in each place the old bundle has a null check and a throw, the new bundle has a call to invariant.

Strangely, further down I do then see a copy of invariant in the old bundle though (one that is **not** included in the new bundle)... So.. You would assume these would cancel each other out no?

Also, all of invariant including the license header is 1.4k, a lot less that the 38k difference we are seeing reported?

Overall, just looking at the diff, if anything I would actually expect the **old** bundle to be bigger... I'll show you the diff when you're around.

.size-snapshot.json
alexreardon: Here is the reason for using invariant:

```js
 * The invariant message will be stripped in production, but the invariant
 * will remain to ensure logic does not differ in production.
```

Flow understands that invariant will throw if the condition is not met

lukebatchelor:
Oh, also worth noting, the actual sizes I'm getting don't match the output above

```
$ cat original.js | wc -c
372696

$ cat new.js | wc -c
372551

$ echo "372696 - 372551" | bc
145
```

Only 145 bytes difference.

package.json
```diff
@@ -56,9 +56,10 @@
   },
   "devDependencies": {
     "@atlaskit/css-reset": "^2.0.0",
-    "@storybook/react": "^3.2.16",
+    "@storybook/react": "^3.4.1",
     "babel-cli": "^6.26.0",
-    "babel-eslint": "^8.2.1",
+    "babel-core": "^6.26.0",
```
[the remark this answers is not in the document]
alexreardon: it was being brought in by another dep. now we are being more explicit

package.json
```diff
@@ -44,21 +44,22 @@
...
     "memoize-one": "^3.1.1",
     "prop-types": "^15.6.0",
     "raf-schd": "^2.1.1",
     "react-motion": "^0.5.2",
     "react-redux": "^5.0.6",
     "redux": "^3.7.2",
     "redux-thunk": "^2.2.0",
-    "reselect": "^3.0.1"
+    "reselect": "^3.0.1",
+    "tiny-invariant": "^0.0.3"
```
alexreardon: right now i have not figured out how to get our rollup build to tree shake the env conditional insight of this module. However, i am happy to solve that later

src/state/dimension.js
```diff
@@ -134,10 +134,11 @@ export const scrollDroppable = (
     clipped,
   };
 
-  return ({
+  const result: DroppableDimension = {
```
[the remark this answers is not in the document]
alexreardon: nope! it now works with exact types and spread

alexreardon:
I have no idea where the extra 27,156 bytes came from... I had a look at the diff and i cannot see how we got such an increase 

Original

```bash
cat dist/react-beautiful-dnd.js | wc -c
  372696
```

New

```bash
cat dist/react-beautiful-dnd.js | wc -c
  399852
```

alexreardon:
Given that it will only impact the umd distribution i am okay to proceed at this point. However, it would be worth getting to the bottom of this one

alexreardon:
This could be relevant:

```
./src/index.js → dist/react-beautiful-dnd.js...

Computed sizes of "dist/react-beautiful-dnd.js" with "umd" format
  bundled: 399,839 B
  minified with uglify: 149,239 B
  minified and gzipped: 42,623 B


(!) Missing exports
https://github.com/rollup/rollup/wiki/Troubleshooting#name-is-not-exported-by-module
commonjs-proxy:/Users/areardon/code/react-beautiful-dnd/node_modules/core-js/library/modules/es6.object.to-string.js
default is not exported by node_modules/core-js/library/modules/es6.object.to-string.js
1: import * as es6_object_toString from "/Users/areardon/code/react-beautiful-dnd/node_modules/core-js/library/modules/es6.object.to-string.js"; export default ( es6_object_toString && es6_object_toString['default'] ) || es6_object_toString;
```

TrySound:
Seems like the problem with size and missing exports is related to new rollup version and core-js.

TrySound:
Yep, it's regression in rollup

alexreardon:
Thanks @TrySound. I will merge this PR as is and we can create a follow up issue to fix the umd build

PR #449, opened 2018-04-18
Maintain focus when moving between lists

alexreardon:
Previously there was a lot of ad hoc code to maintain focus of a draggable when it moved from one list to another. It is so important for the user experience that I have baked it into the library


alexreardon:
This will result in a minor version change

README.md
```diff
@@ -692,15 +692,6 @@ Here are a few poor user experiences that can occur if you change things *during
 - If you remove the node that the user is dragging, then the drag will instantly end
 - If you change the dimension of the dragging node, then other things will not move out of the way at the correct time.
 
-#### Force focus after a transition between lists
```
alexreardon: We now do this out of the box

src/view/draggable/focus-on-drag-handle.js
```diff
@@ -0,0 +1,28 @@
+// @flow
```
alexreardon: A new file to help us to get the draggable ref which is currently not provided by consumers

test/unit/view/unconnected-draggable.spec.js
```diff
@@ -1367,6 +1374,161 @@ describe('Draggable - unconnected', () => {
     });
   });
 
+  describe('Cross list movement focus retention', () => {
```
alexreardon: tests for everyone!

website/src/components/examples/board/board.jsx
```diff
@@ -40,7 +40,6 @@ type Props = {|
 type State = {|
   columns: QuoteMap,
   ordered: string[],
-  autoFocusQuoteId: ?string,
```
alexreardon: I got rid of all the lame custom auto focus code

README.md
```diff
@@ -1027,6 +1018,10 @@ It is a contract of this library that it owns the positioning logic of the dragg
 
 To get around this you can use [`React.Portal`](https://reactjs.org/docs/portals.html). We do not enable this functionality by default as it has performance problems. We have a [using a portal guide](/guides/using-a-portal.md) explaining the performance problem in more detail and how you can set up your own `React.Portal` if you want to.
 
+##### Focus retention when moving between lists
```
alexreardon: This explains the rationale for the change

src/view/data-attributes.js
```diff
@@ -0,0 +1,5 @@
+// @flow
```
alexreardon: rather than repeating these everywhere i pulled them out

src/view/draggable/draggable.jsx
```diff
@@ -75,6 +78,18 @@ export default class Draggable extends Component<Props> {
     this.styleContext = context[styleContextKey];
   }
 
+  componentDidMount() {
+    if (!this.ref) {
+      console.error(`
```
alexreardon: yummy new error message

src/view/draggable/focus-retainer.js
```diff
@@ -0,0 +1,102 @@
...
+type FocusRetainer = {|
+  onDragHandleFocus: (draggableId: DraggableId) => void,
+  onDragHandleBlur: () => void,
+  tryRestoreFocus: (draggableId: DraggableId, draggableRef: HTMLElement) => void,
+|}
+
+// our shared state
+let retainingFocusFor: ?DraggableId = null;
+
+// If we focus on
```
alexreardon: TODO: add comment

test/unit/view/unconnected-draggable.spec.js
alexreardon: these exercise the feature fairly hard

README.md
alexreardon: Given that we are controlling the focus I would like to call it out. I have found that calling behaviour out leads to the least confusion (and issues raised). For example: [how we use dom events](/docs/guides/how-we-use-dom-events]

alexreardon:
I have pushed up a simplification so that we only do the focus retention on a drag end which is way less work on the browser. I'll update the tests tomorrow

README.md
alexreardon: We call out other pieces of information like this through the docs

petegleeson: To me it seems like this change makes the focus behave more intuitively. What do you think about moving this section into the "how we use dom events" page? If you think this is important enough to have in the README I am happy to leave it as is.

src/view/draggable/draggable.jsx
```diff
@@ -188,13 +186,6 @@ export default class Draggable extends Component<Props> {
     // At this point the ref has been changed or initially populated
 
     this.ref = ref;
-
-    // After a ref change we might need to manually force focus onto the ref.
```
alexreardon: focus retention logic is now controlled by the component that has focus: the drag handle

src/view/draggable/draggable.jsx
```diff
@@ -44,7 +44,6 @@ export default class Draggable extends Component<Props> {
   /* eslint-disable react/sort-comp */
   callbacks: DragHandleCallbacks
   styleContext: string
-  isFocused: boolean = false
```
alexreardon: now handled by the drag handle

src/view/drag-handle/drag-handle.jsx
```diff
@@ -71,20 +80,53 @@ export default class DragHandle extends Component<Props> {
...
+    // storing a reference for later
+    this.lastDraggableRef = draggableRef;
+
+    if (!draggableRef) {
+      console.error('Cannot get draggable ref from drag handle');
+      return;
+    }
+
+    // drag handle ref will not be available when not enabled
+    if (!this.props.isEnabled) {
```
Blasz: Unrelated to this PR but any reason why this prop isn't called `isDisabled` instead? It wasn't immediately obvious to me what this prop meant exactly. Changing this to `isDisabled` also aligns with the draggable's disabled prop.

test/unit/view/unconnected-draggable.spec.js
```diff
@@ -266,6 +272,32 @@ const customViewport: Viewport = {
   subject: getArea({ top: 200, left: 100, right: 300, bottom: 300 }),
 };
 
+const looseFocus = (wrapper: ReactWrapper) => {
```
Blasz: s/loose/lose

src/view/drag-handle/drag-handle-types.js
```diff
@@ -54,6 +52,8 @@ export type Props = {|
   isEnabled: boolean,
   // whether the application thinks a drag is occurring
   isDragging: boolean,
+  // whether the application thinks a drop is occurring
+  isDropAnimating: boolean,
```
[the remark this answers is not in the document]
alexreardon: Perhaps! We currently use this naming in multiple places so i i'll leave it for now

src/view/drag-handle/drag-handle.jsx
alexreardon: I generally favour positive props for internal usage. As you mentioned, not part of this PR - but perhaps work thinking more about

src/view/data-attributes.js
```diff
@@ -0,0 +1,5 @@
...
+export const prefix: string = 'data-react-beautiful-dnd';
+export const dragHandle: string = `${prefix}-drag-handle`;
+export const draggable: string = `${prefix}-draggable`;
+export const droppable: string = `${prefix}-droppable`;
```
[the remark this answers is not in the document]
alexreardon: i generally type every variable even if it can be inferred

PR #472, opened 2018-04-26
most of website now has a layout

Noviny:
@alexreardon Got a lot of work done on this, but ran into an issue with codesandboxer not parsing `.jsx` files when I was trying to add it in to the examples pages.

Should probably talk about the examples as well, as they don't work very well in the space this version of the site provides them.

My next step is update codesandboxer, and reconvene with you to talk about examples.


website/gatsby-node.js
```diff
@@ -133,3 +137,19 @@ exports.createPages = ({ graphql, boundActionCreators }/* : NodeParams */)/* : P
     );
   });
 };
+
+exports.onCreatePage = async ({ page, boundActionCreators }) => {
+  const { createPage } = boundActionCreators;
+
+  return new Promise((resolve, reject) => {
```
alexreardon: kewl! Does this still work well with SSR?

alexreardon: Does it need to be a promise?

alexreardon:
I am keen to iterate on this. Let's get the build passing an merge it in

src/state/move-to-next-index/in-home-list.js
```diff
@@ -4,7 +4,6 @@ import getDraggablesInsideDroppable from '../get-draggables-inside-droppable';
 import { patch, subtract } from '../position';
 import withDroppableDisplacement from '../with-droppable-displacement';
 import isTotallyVisibleInNewLocation from './is-totally-visible-in-new-location';
-// import getScrollJumpResult from './get-scroll-jump-result';
```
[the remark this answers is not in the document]
Noviny: Codesandboxer by default was attempting to add this file to the bundle, which was causing a fairly large break. Preferably, this file itself shouldn't be uploaded. Was this as commented out important to maintain?

website/gatsby-node.js
Noviny: This is the recommended way to do this from gatsby's docs. I don't think it should hit SSR, as this is run when gatsby is generating the site's data.

Noviny:
I'm pretty confused. The build errors here are to do with unused variables that I removed on my latest PR.

src/state/move-to-next-index/in-home-list.js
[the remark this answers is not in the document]
alexreardon: no idea why it is just commented..

website/src/components/Sidebar/index.jsx
```diff
@@ -0,0 +1,100 @@
+// @flow
+import React, { Fragment } from 'react';
+import Link from 'gatsby-link';
+import styled, { css } from 'styled-components';
+import { colors as akColors } from '@atlaskit/theme';
+import { colors } from '../../constants';
+import type { docsPage, sitePage } from '../types';
+import { getTitleFromExamplePath } from '../../utils';
+
+const Sidebar = styled.div`
```
alexreardon: I would ask that these be based on the grid, but given that i plan on moving the sidebar out we can leave it like this for now

alexreardon:
Maybe pull in master?

alexreardon:
If that fails, perhaps recreate the pr

PR #477, opened 2018-04-28
bumping redux and others

alexreardon:


flow-typed/npm/redux_v3.x.x.js
```diff
@@ -1,59 +0,0 @@
-// flow-typed signature: cca4916b0213065533df8335c3285a4a
```
alexreardon: sadly there is no v4 yet

PR #486, opened 2018-05-02
fix(types): compatible with csstype

bradleyayers:
Switch from Maybe types to optional properties. This makes the types compatible with `csstype`'s definition.

Currently the `style` object passed to consumers doesn't have a type compatible with React's intrinsic elements' `style` type. In general it's due to using `null` rather than `undefined`.

### DIFF
--- a/src/view/moveable/moveable.jsx
+++ b/src/view/moveable/moveable.jsx
@@ -16,7 +16,7 @@ const origin: Position = {
 };
 
 const noMovement: Style = {
-  transform: null,
+  transform: undefined
 };
 
 const isAtOrigin = (point: PositionLike): boolean =>
--- a/README.md
+++ b/README.md
@@ -986,11 +986,11 @@ type DraggingStyle = {|
   left: number,
   margin: 0,
   transition: 'none',
-  transform: ?string,
+  transform?: string,
   zIndex: ZIndex,
 |}
 type NotDraggingStyle = {|
-  transition: ?string,
+  transition?: string,
   transition: null | 'none',
 |}
 ```
@@ -1342,12 +1342,12 @@ type DraggingStyle = {|
   left: number,
   margin: 0,
   transition: 'none',
-  transform: ?string,
+  transform?: string,
   zIndex: ZIndex,
 |}
 type NotDraggingStyle = {|
-  transition: ?string,
-  transition: null | 'none',
+  transition?: string,
+  transition?: 'none',
 |}
 
 type DragHandleProps = {|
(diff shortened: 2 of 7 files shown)


alexreardon:
I generally prefer optional values over optional properties. But I do not want to get in the way of having correct types also. I thought that having a null property on a style object was fine for react

bradleyayers:
Here's a little more background to it:

https://github.com/frenic/csstype builds [Flow](https://github.com/frenic/csstype/blob/master/index.js.flow) and [TypeScript](https://github.com/frenic/csstype/blob/master/index.d.ts) types from the [MDN data](https://github.com/mdn/data/blob/master/css/properties.schema.json), and [React's TypeScript types now use `csstype`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L24) and [defines `style` as an optional property](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L953). csstype defines [`transform` as an optional property](https://github.com/frenic/csstype/blob/master/index.js.flow#L274) rather than a maybe (which means `null` is not allowed in the type).

src/view/moveable/moveable.jsx
```diff
@@ -16,7 +16,7 @@ const origin: Position = {
 };
 
 const noMovement: Style = {
-  transform: null,
+  transform: undefined
```
alexreardon: is it better just to remove the property?

alexreardon:
This looks good, but i want to have a good play with the branch as our styling can be a bit fiddly. Also, is there anyway we can validate against csstype to ensure we are not breaking it going forward?

src/view/moveable/moveable.jsx
bradleyayers: I had that thought too, but flow complained about `{}` not matching the shape, so I switched to this approach of just using `undefined`.

bradleyayers:
It might be worth spiking depending on csstype and using the definitions from there, rather than having custom definitions in react-beautiful-dnd. Perhaps something like:

```
import { type Properties } from "csstype";

type Style = {
  transform?: $PropertyType<Properties, "transform">;
}
```

alexreardon:
@bradleyayers do you have some time to discuss this today? We are about to push out a release so it would be a good time to address this

alexreardon:
I am going to close this due to inactivity. If somebody feels passionate about it they are welcome to start this one back up.

For now we are conforming correctly to React style api

PR #496, opened 2018-05-09
Add stylelint

### LINKED ISSUE #373: Add stylelint
I think it would be worth adding style linting to this repo. Even if it is only for the examples which use styled components.

Given your recent experience with this @Blasz I have assigned this to you - but feel free to let me know if you are not keen

xxhomey19:
close #373 

Hi! This is an awesome library, and I'm trying to contribute. The following is the instruction of this PR. 
1. `stylelint` only runs with the files in `/stories`, because it seems to be that no other place uses CSS.
2. Since the latest `stylelint` seems to have a [bug](https://github.com/stylelint/stylelint/issues/3252) with [stylelint-processor-styled-components](https://github.com/styled-components/stylelint-processor-styled-components), so I installed `stylelint@9.1.3` for now.
3. Add `/*stylelint disabled*/` in several places to prevent some strange or unnecessary `stylelint` errors.

Please let me know if there are any questions. Thanks!

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,0 +1,7 @@
+{
+  "processors": ["stylelint-processor-styled-components"],
+  "extends": [
+    "stylelint-config-standard",
+    "stylelint-config-styled-components"
+  ]
+}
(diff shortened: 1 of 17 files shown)


alexreardon:
Brilliant!

PR #550, opened 2018-06-15
Fix link and newlines

TrySound:


alexreardon:
Good catch! I did not know about this PR so i already merged a fix. However, my fix did not include the white space removals. Could you update this pr? 😊

Issue #219, opened 2017-12-08
Use keyboard to move between to stacked vertical lists


wschuster-nm:
## Bug or feature request?
Feature Request (possibly related to #40)

### Expected behavior
To be able to move`<Draggable />`s with a keyboard between 2 vertical `<Droppable />`s that are ore aligned 1 above the other and not side by side.
Possibly using shift+arrow or when the item is at the end of a list it attempts to move to a list above/below?

### Actual behavior
The only way to use the keyboard to move items between a list is between 2 vertical lists placed side by side, or 2 horizontal lists placed 1 above the other,

### Demo
https://codesandbox.io/s/6jlyl9rxqk
(https://www.webpackbin.com was broken when I attempted to use it, the preview was unavailable)

alexreardon:
So cool that you raised this @wschuster-nm. I actually wanted to ship this originally when I was working on keyboard transitioning between lists, but i cut it to reduce scope and get things shipped. I think this feature is really important. I am not sure if we need to introduce the `shift` addition - my initial thought is that when you get to the bottom of a list pressing down would simply move the thing to a `Droppable` below if it can.

wschuster-nm:
Awesome, I completely agree with the "move to next `Droppable`" idea. Also, thank you for making such a great component.

illiphilli:
I'd also really like to see this feature. I'm a bit new to React, and this component made it much easier to get DnD up and running pretty quickly. Keyboard Accessibility is very high in my "needs" list, this feature in particular. In my case, I'm replacing an old Flash matching DD component where I have a list of (draggable) options on the left, and stacked droppable on the right. The keyboard dragging, of course, only works with the top drop zone. 

<img width="1160" alt="screenshot 2017-12-12 11 28 05" src="https://user-images.githubusercontent.com/39205/33896279-69f599fe-df30-11e7-8d26-a3e48c709524.png">

<img width="1154" alt="screenshot 2017-12-12 11 36 20" src="https://user-images.githubusercontent.com/39205/33896399-b631d5b2-df30-11e7-8f85-a0dbec194802.png">

illiphilli:
Hello, and thanks again for the wonderful work on this project. Is there any sense of when this issue might get some traction?

jferrogarcia:
How soon is soon(ish)?

ChelseaRyanAnderson:
[...]

Issue #511, opened 2018-05-18
Add combining support (aka grouping, merging)


alexreardon:
Adding the ability to hover a `Draggable` on top of another one. 

- This would change the sliding of items to allow for a period of overlap
- The onDragUpdate > DragUpdate type would need to be updated to include the grouping information
- The onDragEnd > DragResult type would also need to be updated
- keyboard interactions would need to be aware of grouping
- A `Droppable` would need to opt into allowing it's child `Draggables` to be groupable (still need to figure out what the opt in story looks like)

offerijns:
Really interesting idea! I wonder if this would enable the following use case:

Let's say you have a sidebar on the left with a list of `projects`, which can be reordered. On the right side, you have a list of `items` within the currently selected `project`, which can be reordered as well. You can also drag an `item` to the sidebar, hovering over a `project`, at which point it will be moved to that particular `project` when dropped.

I think this would require some kind of `groupableType` prop on `Droppables`. The `Droppable` of the sidebar would be something like `<Droppable type="project" groupableType="item" />`.

What do you think @alexreardon ? Also, note that the Trello link you shared is not working, it shows the following error:

> **Card not found.**
> This card may be on a private board. If someone gave you this link, they may need to invite you to one of their boards or teams.

alexreardon:
Making progress

![grouping-teaser](https://user-images.githubusercontent.com/2182637/44073300-a1ef2a48-9fd6-11e8-81b8-b0f85d936f97.gif)

offerijns:
Looks awesome, @alexreardon !

Would it be possible to add some kind of attribute like `groupableType`, so you can only group items of a certain type on an item? That would be really helpful for the use case I described above.

alexreardon:
![grouping-with-keyboard](https://user-images.githubusercontent.com/2182637/45340494-2438cb80-b5da-11e8-9777-9c83fe9ef828.gif)

Combining with a keyboard 👍

yduman:
I am really looking forward for this feature, since it will be one of the core actions of my project. You said on another issue that this will be landing on version 10. When can we expect this version? Any approximate date?

mpseidel:
Nice one. Looking forward to using it for feature I plan to build.

alexreardon:
Closed by #838. [See version 10 release notes](https://github.com/atlassian/react-beautiful-dnd/releases/tag/v10.0.0)

Issue #484, opened 2018-05-01
Allow dynamic additions / removals of dimensions during a drag


alexreardon:
This would need to work for both Draggable and Droppable dimensions

Needed to facilitate virtualisation #68.

It would also be useful as its own thing. By supporting this we then unlock lazy loading lists during a drag

alexreardon:
On paper

![img_3073](https://user-images.githubusercontent.com/2182637/39457021-35729256-4d2e-11e8-9a11-4d7d6b31d798.JPG)

alexreardon:
On paper:

![img_3075](https://user-images.githubusercontent.com/2182637/39457044-5f9a0c3a-4d2e-11e8-887f-e14ea1e27d6a.JPG)

alexreardon:
Removals is actually a bit more complicated. If we remove a Draggable from a list then we really need to capture the dimensions for everything after the removed item as well as the list itself. Good times

alexreardon:
Here I am thinking things would be easy. I was wrong.

### My troubles

As soon as you introduce dimension changing there are a few things that can happen:

Adding anything can push other items around
- Adding a Draggable to a Droppable can move other Draggables down
- Adding a Draggable to a Droppable can change the size of the Droppable
- Adding a Draggable to a Droppable can push other Droppables and its children Draggables around
- Adding a Droppable can push other Droppable's around, and its children Draggables

Removing anything can push other items around for similar reasons

### Option 1: Saftey

The only safe option is to recapture all dimensions when anything is added or removed. This feels super heavy. Thankfully we already have an optimised pipeline for this. We can also batch many updates into a single collection phase: eg if you add 5 Draggable's at a time then this would be just one recapture phase.

Additionally, we would need to improve how we handle cancel animations. Currently this is done by animating back to the original position. However, that may no longer be the actual position. Therefore we will need to use our exisiting getHomeOffset function to calculate the drop location. This is not that bad as we already do this when dropping in any other location.

### Option 2: restrictions

We could have a more optimised solution if we enforce a number of limitations on the experience. I am not totally sure what they would all be yet. Things like:

- You cannot add dimensions that would impact the natural starting position of a Draggable
- You cannot add ... WIP

alexreardon:
What we need for virtualisation:

- to capture elements as they are added
- allow dimensions to be removed
- We might need to account for the change in droppable scroll at the time of the mount of a draggable

What we do not need to do:

- recompute the dimensions of the list
[...]

Issue #77, opened 2017-09-07
Animated placeholders when transitioning between lists


alexreardon:
When transitioning between lists there is some snapping that occurs:

- when entering a new list the placeholder snaps to the correct height without animating
- when dropping in a foreign list, the original placeholder in the home list does not animate and disappears instantly when the drop completes

The goal is to have both of these scenarios animated. However, in order to ship #2 it was decided that this feature was not critical enough to hold up the release. The interaction still feels really nice - it would just be even better with animated placeholders.

souporserious:
I'd love to help out here as well if possible. If you have any sort of roadmap or ideas for this I can start a PR.

souporserious:
I landed on this implementation for now which somewhat works, just need to update the dimensions snapshot somehow:

```
function SortableItem({ id, index, children }) {
  return (
    <SortableContainerContext.Consumer>
      {sortableContainerSnapshot => (
        <Draggable draggableId={id} index={index}>
          {(provided, snapshot) => (
            <div>
              <div
                ref={provided.innerRef}
                {...provided.draggableProps}
                {...provided.dragHandleProps}
                style={getItemStyle(
                  snapshot.isDragging,
                  provided.draggableProps.style
                )}
              >
                {children}
              </div>
              {provided.placeholder && (
                <div
                  style={{
                    height: sortableContainerSnapshot.isDraggingOver ? 51 : 0,
                  }}
                >
                  {provided.placeholder}
                </div>
              )}
            </div>
          )}
        </Draggable>
      )}
    </SortableContainerContext.Consumer>
  )
}
```

@alexreardon is there anything exposed right now that would allow me to fire a recalculation of dimensions?

souporserious:
Needs some internal work to clean things up, as well as use React Motion, but pretty cool interaction! 

![animate-on-leave](https://user-images.githubusercontent.com/2762082/36129929-9ea10552-101e-11e8-92c6-8f1f10ab721d.gif)

yogaboll:
I would love to see this fixed. Any progress on this issue?
I can't get @souporserious workaround to function properly since my list-items have differing heights.

souporserious:
@yogaboll you should be able to measure the DOM node and provide that to the height of a wrapper around `provided.placeholder`. The only thing is the drag and drop get's thrown off because the snapshot of the DOM is out of sync with react-beautiful-dnd. Not sure if there is a way to tell it to update that snapshot after the animation completes.

[...]

PR #555, opened 2018-06-19
Use keyboard to move item between vertically stacked lists

sbutkaliuk:
This adds possibility to move item to the next list (vertically) with same UP/DOWN buttons that used to move through the list itself. Fixes: https://github.com/atlassian/react-beautiful-dnd/issues/219

Solution is based on existing movement between different lists but adds opposite axis parameter.

Testing: add `flex-direction: column` to multi-drag/task-app.jsx#Container

### DIFF
--- a/src/state/axis.js
+++ b/src/state/axis.js
@@ -1,5 +1,5 @@
 // @flow
-import type { HorizontalAxis, VerticalAxis } from '../types';
+import type { Axis, HorizontalAxis, VerticalAxis } from '../types';
 
 export const vertical: VerticalAxis = {
   direction: 'vertical',
@@ -24,3 +24,10 @@ export const horizontal: HorizontalAxis = {
   crossAxisEnd: 'bottom',
   crossAxisSize: 'height',
 };
+
+export function oppositeAxis(axis: Axis) {
+  if (axis.direction === 'horizontal') {
+    return vertical;
+  }
+  return horizontal;
+}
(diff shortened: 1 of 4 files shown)


nehaabrol87:
@alexreardon  This is something we need for our app.
I want to check did we follow protocol for opening this PR?
If not, who do we have to alert.

alexreardon:
Very cool!!! 🎉 I am keen to see this happen. I am hesitant to make this change until #77 (and there is another placeholder issue too). 

That will give us a better base before adding this. It should happen not to long from now as #219 is a priority for us. Once #484 and #511 are done we will start working on the placeholders which will unblock this. We should also build a good example that exercises this well with droppable and page scrolling included.

Sorry to hold things up, but i don't want to lock us into a corner just yet!

alexreardon:
I am really keen for this - i think we should just wait a little bit longer for some ground work. We might just be able to reopen this and go from there!

nehaabrol87:
Thanks @alexreardon  for now we will fork the repo.
Any idea around what timeframe would this be incorporated so that we can include a TODO in our code based and we can stop using the forked version of this repo.

nehaabrol87:
@alexreardon qq I see for accessibility support it only says single list?
Is that true?
https://react-beautiful-dnd.netlify.com/?selectedKind=Accessibility&selectedStory=single%20list&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

We are trying to implement it for multiple vertical lists and were not sure if this library is a good fit though.

alexreardon:
It works perfectly for multiple lists. 

Here is our screen reader guide: https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/guides/screen-reader.md

Issue #456, opened 2018-04-19
ETA on dragging to nested droppables


BennievanderWel:
As I understand, dropping a draggable from a droppable to another droppable with is a sibling of the draggable, is not supported. 

Is this correct? And if so, is it on the roadmap?

Other than that, great work on the package!

alexreardon:
It is not on our immediate roadmap but we hope to support it at some point. We do not give out any dates for any feature - things will ship when they are ready and when we are happy with the level of polish. 

> great work on the package!

Thanks! 😊

lxcid:
Look forward to having this functionality! Just for the record 😁

Issue #301, opened 2018-01-29
Dragging items from parent list into a child list


jaredt67:
## Bug or feature request?
feature request
#
from wiki
```
Independent nested lists - a list can be a child of another list, but you cannot drag items from the parent list into a child list
```

Are you guys planning on adding the functionality of dragging items from parent list into a child list in the future? Thank you!

alexreardon:
At this time it is not on our roadmap. However, once a few more features land we will consider it!

kriim:
This would be awesome.

`react-beatiful-dnd` would be a perfect fit if we could move items between parent and child lists (both directions and with multiple levels).

Would this be hard to add as a feature?

skellystudios:
@alexreardon I'd be keen to have a go making this possible.

Are there any large architectural barriers you can think of to being able to do that? Do you have any high-level pointers for what would be required?

alexreardon:
I have not looked into the problem so i am not 100% across all the things you would need to look into.

Off the top of my head:

- when getting the droppable over you would need to check for the smallest droppable you are over rather than the first
- There is an assumption that a droppable cannot move during a drag. This assumption would be broken if the inner list was in a foreign list
- The keyboard story would need to be thought about. How would a use know that the next item is a nested container and how would they move into it? Would we move into it by default or skip it?

There are probably others but that is my initial thoughts!

mattkrick:
@skellystudios any progress on this? i'd be happy to team up on this one

skellystudios:
Hey Matt –

So I made a bit of a start, and got it working in its most basic form – check out here: https://github.com/skellystudios/react-beautiful-dnd

Unfortunately, I got a bit unstuck working on the keyboard controls, and trying to get it tested in any reasonable way, and I've been putting that off. 

If you're able to get it into a mergeable state, that would be amazing

mattkrick:
thanks!
yeah, building it to expose the droppableId to parents & children isn't an easy fix. i took a different approach & just stacked multiple vertical lists on top of each other. It still has the keyboard problem, but I saw #219 has a priority so it may get support sooner than "true" nesting.

Issue #134, opened 2017-10-05
Dragging Items Between Nested Lists


kgorgi:
## Bug or feature request?
Question/Feature Request

In one of your provided examples, [nested vertical list](https://react-beautiful-dnd.netlify.com/?selectedKind=complex%20vertical%20list&selectedStory=nested%20vertical%20lists&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel), a user can drag within a nested list but not drag between nested lists (dragging an item from _second level_ to the _top level_ or vice versa in the linked example above). Is this use-case currently possible with react-beautiful-dnd, dragging items between nested lists?

kgorgi:
I just found in the README.md that the library does not support this (see below). Thus answering my own question. 

> Independent nested lists - a list can be a child of another list, but you cannot drag items from the parent list into a child list

So let me change my question: Is there are plans for this type of feature in the future?

jaredcrowe:
hey @kgorgi, this is definitely something which we'd be open to adding in the future, but it's not on our immediate roadmap. right now we're focused on nailing the experience for sorting flat lists :)

williaster:
we have a huge need for this, any pointers on direction / would you accept a PR for this?

williaster:
@jaredcrowe wanted to bump my ^question

jaredcrowe:
Hey @williaster this isn't on our roadmap right now, and at this stage I'd say we'd consider it out of scope for this library.

ajay-gandhi:
@williaster You might be able to use the technique mentioned in #212

krisl:
is it supported to drag items between two sibling lists? eg drag item between two second level containers?

PR #563, opened 2018-06-21
Fix parent to child drops

ianstormtaylor:
This lets draggable items be dragged from a parent and dropped into a nested child list, making this library suitable for not just boards but also trees.

This was previously impossible, since the parent list would only ever register itself, but by dropping into the smallest list instead, we can drop into children too.

Right now this change works, in that you can now drag and drop either from a child into a parent, or from a parent into a child.

But there's one thing missing, which is that when dragging between nested lists, when a placeholder is rendered in the non-home list, for some reason it doesn't also increase the list's height to account for the placeholder. You can see this behavior here:

@alexreardon I couldn't for the life of me figure out where this logic was in the codebase. Is there any way this is something you could finish? or point me to where I need to make a change?

Thank you for this great library!

Fixes https://github.com/atlassian/react-beautiful-dnd/issues/456
Fixes https://github.com/atlassian/react-beautiful-dnd/issues/301
Fixes https://github.com/atlassian/react-beautiful-dnd/issues/134

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,21 +1,21 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 400269,
-    "minified": 151652,
-    "gzipped": 43139
+    "bundled": 400507,
+    "minified": 151621,
+    "gzipped": 43147
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 358556,
-    "minified": 134751,
-    "gzipped": 38201
+    "bundled": 358937,
+    "minified": 134847,
+    "gzipped": 38228
   },
   "dist/react-beautiful-dnd.esm.js": {
-    "bundled": 177759,
-    "minified": 89807,
-    "gzipped": 22649,
+    "bundled": 177889,
+    "minified": 89747,
+    "gzipped": 22655,
     "treeshaked": {
-      "rollup": 78288,
-      "webpack": 80060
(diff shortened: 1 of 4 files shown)


ianstormtaylor:
@alexreardon actually, false alarm! It was a simple fix, the storybook was just accidentally not rendering the `dropProvided.placeholder` in the list.

So now dragging from a parent to a child (or a child to a parent) works!

alexreardon:
Hi @ianstormtaylor,

Thanks for

alexreardon:
Hi @ianstormtaylor, 

thanks for putting this forward. It is a really interesting addition! However, at this point I do not think it best to proceed with this PR. 

Rationale:

- We have made some major refactors #493 which may invalidate this change a little bit
- I am hoping to do some really big work to make placeholders work nicer #77 (I think there is another issue too). This change will impact this work. Until that lands I am keen to avoid adding changes like this for now

You're a legend for raising this. Hopefully we can look into it after #77 lands!!

ianstormtaylor:
Hey @alexreardon any thoughts on this? 

I was just thinking about this more, I feel like this is the best way to get folks like myself and others from #134, #301, #456, etc. being able to using this functionality so that we can submit fixes/improvements as we use it. Otherwise it's a Catch 22 of never being able to use it, and then not finding the deeper issues. (I really think there are many fewer issues than we think there will be.)

Thank you!

PR #591, opened 2018-06-28
Fixing droppable keyboard scroll

alexreardon:


src/state/reducer.js
```diff
@@ -297,6 +297,8 @@ export default (state: State = idle, action: Action): State => {
       phase: state.phase,
       impact,
       dimensions,
+      // At this point any scroll jump request would need to be cleared
+      scrollJumpRequest: null,
```
alexreardon: this is the fix

PR #597, opened 2018-06-29
Replace invariant with shorter version in production

TrySound:
4kb less to parse, 1kb less to load in production

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -58,12 +58,11 @@ export default [
     // Only deep dependency required is React
     external: ['react'],
     plugins: [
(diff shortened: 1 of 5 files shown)


rollup.config.js
```diff
@@ -58,12 +58,11 @@ export default [
     // Only deep dependency required is React
     external: ['react'],
     plugins: [
-      // Setting production env before running babel etc
-      replace({ 'process.env.NODE_ENV': JSON.stringify('production') }),
```
alexreardon: Why did this need to move down?

alexreardon:
🤘

PR #605, opened 2018-07-02
Bumping rollup snapshot. Fixes #483

alexreardon:
/cc @TrySound 
Fixes #483


alexreardon:
Adding the bundle size check back into the travis build

rollup.config.js
```diff
@@ -22,7 +22,18 @@ const getBabelOptions = ({ useESModules }) => ({
...
+const snapshotArgs = (() => {
+  const shouldMatch = process.env.SNAPSHOT === 'match';
+
+  if (!shouldMatch) {
+    return {};
+  }
+
+  return {
+    matchSnapshot: true,
+    threshold: 1000,
```
TrySound: `threshold` could be just passed. It's ignored with `matchSnapshot === false`.

alexreardon: I am happy to keep it as is because it is clear - but you're more than welcome to change it 👍

PR #610, opened 2018-07-02
Disabling dynamic changes

alexreardon:
Disabling the uncompleted feature for 8.0 #484

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 359505,
-    "minified": 135309,
-    "gzipped": 37912
+    "bundled": 351441,
+    "minified": 131710,
+    "gzipped": 36933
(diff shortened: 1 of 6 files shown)


alexreardon:
We can add this back in the next feature release

.size-snapshot.json
```diff
@@ -1,25 +1,25 @@
...
+    "minified": 131710,
+    "gzipped": 36933
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 319661,
-    "minified": 117855,
-    "gzipped": 32234
+    "bundled": 312114,
+    "minified": 114558,
+    "gzipped": 31398
```
alexreardon: a small gain for the release @TrySound

alexreardon: 1kb gzip. I could go lower by removing more unused code for this feature but i won't as we will be adding it right back in 8.1

PR #618, opened 2018-07-04
hook functions are now called after the reducer

### LINKED ISSUE #617: Error when updating state inside onDragStart
I'm calling setState inside onDragStart and I'm getting the following error:

`Error: Invariant failed: Attempting to move in an unsupported phase PREPARING`

16.3.2

8.0.1

Chrome

alexreardon:
Fixes #617

### DIFF
--- a/src/state/middleware/hooks.js
+++ b/src/state/middleware/hooks.js
@@ -239,10 +239,12 @@ export default (getHooks: () => Hooks, announce: Announce) => {
   return (store: Store) => (next: Action => mixed) => (
     action: Action,
   ): mixed => {
+    // letting the reducer update first
+    next(action);
+
     if (action.type === 'INITIAL_PUBLISH') {
       const critical: Critical = action.payload.critical;
       publisher.start(critical);
-      next(action);
       return;
     }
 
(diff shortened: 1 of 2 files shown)


src/view/draggable/draggable.jsx
```diff
@@ -115,15 +117,6 @@ export default class Draggable extends Component<Props> {
     this.ref = null;
   }
 
-  // This should already be handled gracefully in DragHandle.
```
alexreardon: this is already taken care of inside of the drag-handle and the reducer

alexreardon:
This brings the behaviour back in line with previous versions

alexreardon:
Super helpful: https://allyjs.io/tutorials/focusing-in-svg.html

alexreardon:
IE11 will not support .focus() on an SVG. This means if you are dragging an SVG in ie11:

- portals will not work (we need to restore focus after the element moves to the portal)
- focus will not be retained when moving between lists

Right now both of these will throw an exception in IE11 (same for 7.1.3)

I am not sure what the best path forward is. We could either make a note of it and provide some tips, or we could continue to block the use of SVG's as drag handles as we are now... 🤔

PR #624, opened 2018-07-06
Upgrade babel to 52 and upgrade rollup

TrySound:
All works fine


alexreardon:
Thanks for picking this up @TrySound. I'll leave the babel / rollup bumps for you while we are on beta. Should we do another release for this? I think we are okay looking at the bundles

TrySound:
There's nothing changed actually. Just deprecations. Different sizes probably because of different yarn versions. Which do you have? Mine is 1.7.0

alexreardon:
The snapshot might be a touch out of date

PR #630, opened 2018-07-09
test: enable puppeteer testing using xfvb to avoid headless flakes

MarshallOfSound:


.travis.yml
```diff
@@ -29,4 +29,8 @@ before_script:
   - greenkeeper-lockfile-update
   - yarn global add wait-port@0.2.2
   - yarn storybook & wait-port -t 10000 localhost:9002
+  - "export DISPLAY=:99.0"
```
alexreardon: Hey @MarshallOfSound, could you add a comment above this line to explain what these two commands are for? Cheers!

PR #632, opened 2018-07-10
Accounting for cross axis scrollbars in auto scrolling. Fixes #628 

### LINKED ISSUE #628: When user in the bottom of scrollable list, scroll up is not working
Bug: When user in the bottom of scrollable list, and drag bottom item to the top, scroll top is not working.

Expected behaviour: List should scroll up when user move draggable item to the top of the list

Actual behaviour: List is not scrollable

Steps to reproduce: Basic example from library with updates: https://codesandbox.io/s/3457zo2kn1

What version of `React` are you using: 16
What version of `react-beautiful-dnd` are you running: latest
What browser are you using: Chrome
Operating system: OSX (mac)

Demo ezgif com-video-to-gif

alexreardon:
Fixes #628

### DIFF
--- a/src/state/auto-scroller/can-scroll.js
+++ b/src/state/auto-scroller/can-scroll.js
@@ -59,6 +59,7 @@ export const canPartiallyScroll = ({
   // It is possible for the max scroll to be greater than the current scroll
   // when there are scrollbars on the cross axis. We adjust for this by
   // increasing the max scroll point if needed
+  // This will allow movements backwards even if the current scroll is greater than the max scroll
   const max: Position = {
     x: Math.max(current.x, rawMax.x),
     y: Math.max(current.y, rawMax.y),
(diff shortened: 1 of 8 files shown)


src/state/auto-scroller/can-scroll.js
```diff
@@ -52,10 +52,19 @@ export const getOverlap = (() => {
...
+  max: rawMax,
   current,
   change,
-}: CanScrollArgs): boolean => {
+}: CanPartiallyScrollArgs): boolean => {
+  // It is possible for the max scroll to be greater than the current scroll
+  // when there are scrollbars on the cross axis. We adjust for this by
+  // increasing the max scroll point if needed
+  // This will allow movements backwards even if the current scroll is greater than the max scroll
+  const max: Position = {
```
alexreardon: this is the fix

PR #633, opened 2018-07-11
fixing invariant check in reducer. Fixes #617

### LINKED ISSUE #617: Error when updating state inside onDragStart

alexreardon:
Fixes #617

### DIFF
--- a/src/state/reducer.js
+++ b/src/state/reducer.js
@@ -332,10 +332,9 @@ export default (state: State = idle, action: Action): State => {
     }
 
     invariant(
-      target.isEnabled === isEnabled,
-      `Trying to set droppable isEnabled to ${String(
-        isEnabled,
-      )} but it is already ${String(isEnabled)}`,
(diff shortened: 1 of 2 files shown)


src/state/reducer.js
```diff
@@ -332,10 +332,9 @@ export default (state: State = idle, action: Action): State => {
     }
 
     invariant(
-      target.isEnabled === isEnabled,
-      `Trying to set droppable isEnabled to ${String(
-        isEnabled,
-      )} but it is already ${String(isEnabled)}`,
+      target.isEnabled !== isEnabled,
```
alexreardon: this invariant check was the wrong way around.

We need to throw if they are the same.

alexreardon: We need these values to be different - otherwise it is an error

PR #200, opened 2017-11-27
Placeholder cleanup

alexreardon:
Closes #158

PR #636, opened 2018-07-11
allow dragging children to/from parent lists

ianstormtaylor:
This lets draggable children be dragged from or into a parent, making this library suitable for not just boards but also trees.

@alexreardon one of the reasons that #563 was closed so quickly was because of the big refactor, so I've rebased this pull request against the latest master. It's still just a ~10 line change that only impacts the nested lists use case.

Would you consider merging so we can unblock people using nested lists?

### DIFF
--- a/src/state/get-droppable-over.js
+++ b/src/state/get-droppable-over.js
@@ -154,7 +154,7 @@ export default ({
   const maybe: ?DroppableDimension = toDroppableList(droppables)
     // only want enabled droppables
     .filter((droppable: DroppableDimension) => droppable.isEnabled)
-    .find(
+    .filter(
       (droppable: DroppableDimension): boolean => {
         // If previously dragging over a droppable we give it a
         // bit of room on the subsequent drags so that user and move
@@ -176,7 +176,25 @@ export default ({
         // if it is over the droppable - not its internal impact
         return isPositionInFrame(withPlaceholder)(target);
       },
-    );
+    )
+    .sort(
+      (a, b): number => {
+        if (
+          a.client.contentBox[a.axis.size] < b.client.contentBox[b.axis.size]
+        ) {
+          return -1;
+        }
+
+        if (
+          a.client.contentBox[a.axis.size] > b.client.contentBox[b.axis.size]
+        ) {
+          return 1;
+        }
+
+        return 0;
+      },
+    )
+    .find((droppable: DroppableDimension): boolean => !!droppable);
 
   return maybe ? maybe.descriptor.id : null;
 };
--- a/stories/src/vertical-nested/quote-app.jsx
+++ b/stories/src/vertical-nested/quote-app.jsx
@@ -61,64 +61,78 @@ export default class QuoteApp extends Component<*, State> {
     publishOnDragEnd(result);
 
     // dropped outside the list
-    if (!result.destination) {
-      return;
-    }
-
-    if (result.type === 'first-level') {
-      const children = reorder(
-        this.state.list.children,
-        result.source.index,
-        result.destination.index,
+    if (result.destination) {
+      const destId: string = result.destination.droppableId;
+      const srcId: string = result.source.droppableId;
+      const list: NestedQuoteList = this.state.list;
+      const nestedList: ?NestedQuoteList = list.children.find(
+        item => item.children,
       );
 
-      // $ExpectError - using spread
-      const list: NestedQuoteList = {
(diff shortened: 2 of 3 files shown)


src/state/get-droppable-over.js
```diff
@@ -176,7 +176,25 @@ export default ({
         // if it is over the droppable - not its internal impact
         return isPositionInFrame(withPlaceholder)(target);
       },
-    );
+    )
+    .sort(
```
alexreardon: Note to self: this is the core change here

alexreardon:
### Short answer

We cannot merge this now

### Long answer

We are hoping to do some work very soon to improve the placeholder experience for the library #200. I want to make sure we get everything stable with that experience before we increase our feature set. 

Additionally, when it comes to nested lists I want to do some more thinking about the keyboard interaction.

### How you can proceed

It is a totally valid option to fork this library and add your desired behaviour into that for the time being. That way you can use the feature even though we do not support it. You will just need to remember to pull in the latest code from time to time to take advantage of our improvements

Karmalakas:
I believe nothing is merged to work with nested lists, right? But still I managed to make it work without any extra code. My structure is like this (blue - droppable; red - draggable):
image

Although there are three main issues with nesting:

1. By default, if draggable has a droppable inside it and, when you drag over the area, where that droppable was, it tries to add placeholder in it (changing the inner droppable height, but leaving the draggable height unchanged) :) you have to manually make that inner droppable disabled
image

2. While draggable is dragged in the same droppable, all is good and the space where original item was, is collapsed. But as soon as you drag it outside to another droppable (no matter if parent or sibling's inner droppable), there are two placeholders visible - one as it should be, beneath the dragable and the other appears in the old position where draggable was originali before moving it (you can see the same in an animation above)
image

3. This is closely related to the 2nd issue. If I try to drag an item outside of it's droppable into another, it doesn't add a placeholder until I drag to the point, where the droppable was with the original item in it's original position. I tried to show it as best as I could :D 
image

Karmalakas:
I'll see what I can do when I'm back to work from sick leave some time next week. If I'll remember :D

Venryx:
@alexreardon Any possibility of this pull request (or a new one based on the same idea) being considered for inclusion, presumably now that the placeholder refactor is complete?

(I'm another user with nested-lists/trees in his project, and would love to be able to switch to this library from react-dnd for the nice animations and simpler integration approach it provides.)

Venryx:
As it doesn't seem like it'll be merged for now, I thought I'd share an easier way to implement the changes above without maintaining your own fork.

Just use the string-replace-webpack-plugin module!: https://github.com/jamesandersen/string-replace-webpack-plugin

It's a great tool that I always include for my large projects. What it lets you do, is specify a block of code to find, and a block of code to replace it with. It works for any file that you require from webpack, and this _includes_ files in the `node_modules` folder.

Specifically, to implement the patch above, here is the code I have added to my webpack config file:
```
webpackConfig.module.rules.push({
	loader: StringReplacePlugin.replace({
		replacements: [
			// make lib support nested-lists better (from: https://github.com/atlassian/react-beautiful-dnd/pull/636)
			{
				pattern: /var getDroppableOver = (.|\n)+?(?=var getDraggablesInsideDroppable)/,
				replacement: () => {
					return `
						var getDroppableOver = (function(args) {
							var target = args.target, droppables = args.droppables;
							var maybe = toDroppableList(droppables)
								.filter(droppable => {
									if (!droppable.isEnabled) return false;
									var active = droppable.subject.active;
									if (!active) return false;
									return isPositionInFrame(active)(target);
								})
								.sort((a, b) => {
									// if draggable is over two lists, and one's not as tall, have it prioritize the list that's not as tall
									if (a.client.contentBox[a.axis.size] < b.client.contentBox[b.axis.size]) return -1;
									if (a.client.contentBox[a.axis.size] > b.client.contentBox[b.axis.size]) return 1;
									return 0;
								})
								.find(droppable => !!droppable);
							return maybe ? maybe.descriptor.id : null;
						});
					`.trim();
				},
			}
		]
	}),
});
```

Done! Your modification is now saved within your project, without needing to create your own variant of the package on npm.

Granted, the regex used to replace the code is not guaranteed to match future versions, so if you take this approach, I suggest locking the version of `react-beautiful-dnd` to the exact version you have installed, and include a warning in your readme telling about the modification. (so other members don't upgrade the library and break the regex-based modification without realizing it)

PR #639, opened 2018-07-16
Updated README.md with valid sandbox url

rambabusaravanan:
"Simple DnD between two lists" is updated with valid sandbox url


rambabusaravanan:
@alexreardon I played with my [fork](https://codesandbox.io/s/j2krvxwpv9) of yours with updated dependency. But the one updated in readme is yours.

Readme has vertical and horizontal list example from your codesandbox profile. So I just searched in your codesandbox profile for 'two list' and updated.

PR #646, opened 2018-07-17
Fix/ie 11 scroll listener

alexreardon:


src/view/droppable-dimension-publisher/droppable-dimension-publisher.jsx
```diff
@@ -184,6 +192,7 @@ export default class DroppableDimensionPublisher extends Component<Props> {
     watching.closestScrollable.removeEventListener(
       'scroll',
       this.onClosestScroll,
+      listenerOptions,
```
alexreardon: this is the ie11 fix - need to remember to call the removeEventListener with the same listener options to remove to the event handler

PR #661, opened 2018-07-24
Rbellebon/add test cafe

RajaBellebon:
Hello @alexreardon ,

I evaluated Cypress and I will write a blog about why I decided to switch to TestCafe.

[Documentation](https://devexpress.github.io/testcafe/documentation/getting-started/)

Regarding the use case of drag and drop and not using selenium or puppeteer, this test framework seems more than enough.

**Pros:**
* Ability to perform [drag](http://devexpress.github.io/testcafe/documentation/test-api/actions/drag-element.html) (using drag function)
* Ability to use the keyboard
* Ability to run on different [browsers](https://devexpress.github.io/testcafe/documentation/using-testcafe/common-concepts/browsers/browser-support.html)
* Ability to run headless
* Ability to run on browserstack (Edge, IE11, access to mobile emulator)

**To be confirmed:**
* Test Executions seem fast
* Seems scalable (meaning we can add more tests)

**Cons:**
* Use of the testCafe runner instead of Jest
* testCafe has a local runner but it is not the best to debug( compare to cypress)

At the moment, the test will run against:
- Chrome headless
- Chrome headless with emulator device for iphone (to validate touch event)
- Firefox headless


test/browser/simple-list.js
```diff
@@ -0,0 +1,41 @@
...
+fixture`Simple List Test Case`.page`${urlSimpleList}`.beforeEach(async t => {
+  await t
+    .navigateTo(urlSimpleList)
+    .expect(Selector(singleListContainer).visible)
+    .ok();
+});
+
+test('Drag and drop using click / drag', async t => {
+  const textBefore = await Selector(secondCard).innerText;
+  await t.dragToElement(secondCard, fourthCard, { speed: 0.5 }).wait(500);
```
alexreardon: wonderful

PR #665, opened 2018-07-27
ci: ensure the yarn lock file does not change during install

MarshallOfSound:


alexreardon:
Close 😊

MarshallOfSound:
@alexreardon All green 👍

PR #681, opened 2018-08-02
bumping dev react and loosing restriction on prop-types

alexreardon:


package.json
```diff
@@ -53,7 +53,7 @@
     "@babel/runtime": "7.0.0-beta.54",
     "css-box-model": "^1.0.0",
     "memoize-one": "^4.0.0",
-    "prop-types": "15.6.1",
+    "prop-types": "^15.6.1",
```
alexreardon: I want to use 15.6.1 internally as 15.6.2 has a big kb blowout. But this change allows consumers to not have to double include prop-types. This will go out in the next release

PR #687, opened 2018-08-05
Upgrade to babel beta 56

TrySound:
Babel runtime was split with @babel/runtime and @babel/runtime-corejs2
pacakges which allows to omit large core-js pacakge in many cases. Babel
runtime with only helpers became a default so users could include
necessary polyfills by themselves.


alexreardon:
What polyfills would be required? I think it worthwhile having a page in
docs/ to explain this

TrySound:
Currently I left as it is. Babel adds core-js `Object.assign` and `Object.keys` polyfills + helper versions with polyfills for es3. So bundle size is mostly increased unnecessary.

TrySound:
Here's the diff after removing core-js
```diff
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 351336,
-    "minified": 131838,
-    "gzipped": 37030
+    "bundled": 346633,
+    "minified": 126400,
+    "gzipped": 35020
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 311485,
-    "minified": 114193,
-    "gzipped": 31352
+    "bundled": 305938,
+    "minified": 108748,
+    "gzipped": 29298
   },
   "dist/react-beautiful-dnd.esm.js": {
-    "bundled": 180986,
-    "minified": 93116,
-    "gzipped": 23145,
+    "bundled": 180804,
+    "minified": 92948,
+    "gzipped": 23104,
     "treeshaked": {
       "rollup": {
-        "code": 68441,
-        "import_statements": 700
+        "code": 68400,
+        "import_statements": 566
       },
       "webpack": {
-        "code": 70717
+        "code": 70575
       }
     }
   }
```

alexreardon:
Do you think it worth dropping core-js? Given it is only a small saving I think we could keep it for now. Thoughts?

PR #720, opened 2018-08-19
Fixing memory leak in DragDropContext. Fixes #715

### LINKED ISSUE #715: Memory leak & onFatalError in DragDropContext
Hello. I started to check memory leaks in our application, and found a lot of them :( 
One of them relies on react-beautiful-dnd. The leak is in the DragDropContext. Let's dive in:

- We have class-component `<DragDropContext/>`
- This class contains: `onFatalError = (error: Error) => {...}`
- and `onWindowError = (error: Error) => this.onFatalError(error)`
- and in `componentDidMount` method we make a subscription: `window.addEventListener('error', this.onWindowError)`
- ... in `componentWillUnmount` - `window.addEventListener('error', this.onWindowError);`

What's wrong with it? These subscriptions retain in memory `<DragDropContext/>` instances forever. And all their links too, of course. They aren't removed from the memory because of `onWindowError` and `onFatalError`. Both of these methods are declared in the class-scope and have link to `this` object. So we have memory leak.

img_17_08_01_049a068531d
^^^ a lot of them :)

img_17_08_01_8efde6c043c

I don't know why DragDropContext needs these subscriptions, but it would be much better to remove them, or rewrite this code without memory leaks ;)

alexreardon:
Fixes #715


src/view/drag-drop-context/drag-drop-context.jsx
```diff
@@ -167,7 +167,7 @@ export default class DragDropContext extends React.Component<Props> {
   }
 
   componentWillUnmount() {
-    window.addEventListener('error', this.onWindowError);
```
alexreardon: Lol! 😊

test/unit/view/drag-drop-context/unmount.spec.js
```diff
@@ -0,0 +1,35 @@
...
+  const wrapper = mount(
+    <DragDropContext onDragEnd={() => {}}>
+      <App />
+    </DragDropContext>,
+  );
+
+  expect(() => wrapper.unmount()).not.toThrow();
+});
+
+it('should clean up any window event handlers', () => {
```
alexreardon: adding a test to ensure window handlers are cleaned up on unmount

PR #722, opened 2018-08-21
Use emotion instead of styled components

Noviny:
I pushed this up yesterday and forgot to make a PR....


alexreardon:
Yum. I'll try this out on my local

alexreardon:
I had a check and it looks like we get some small wins in our examples. Thanks @Noviny !

alexreardon:
I see no reason why we should not put this straight onto master

PR #738, opened 2018-08-23
fixing colours

alexreardon:


website/src/components/common-page.jsx
```diff
@@ -12,7 +12,7 @@ type Props = {
 
 const BaseStyles = styled.div`
   color: ${colors.dark100};
-  background: ${colors.dark500};
+  background: ${colors.dark400};
```
alexreardon: now using dark400 as the base and components will need to opt into dark500 if required

PR #739, opened 2018-08-24
add edit link

Noviny:


website/src/templates/markdown.jsx
```diff
@@ -18,6 +21,7 @@ type Props = {
 
 export default ({ data, location }: Props) => (
   <Layout location={location}>
+    <a href={data.markdownRemark.fields.gitUrl}>Edit these docs</a>
```
alexreardon: Can you please add the edit button to the bottom of the page?

What do you think?

PR #760, opened 2018-08-26
Bcon/fixing codesandboxer

Noviny:
Had some fun with this one. Some of the set-up for codesandbox was kind of hostile to understanding what it was doing, which I'm trying to make a bit nicer. I think ultimately I probably want a `gatsby-codesandboxer` plugin, as we could actually cut out referencing github using gatsby's runtime buuut we probably want a working website in the meantime.

Major changes:
- moved emotion to use emotion syntax (instead of fussing over adding a babelrc to our sandboxes)
- isolated example in its own file so it can be opened in codesandbox (previously it was trying to add the whole page
- provided the `branch` option to codesandboxer's git info so that it is easier to figure discover where codesandboxer is getting the files from

The isolated example pattern is a bit gross, but works for now, and the basic example demonstrates how it works okay.

### DIFF
--- a/.storybook/.babelrc
+++ b/.storybook/.babelrc
@@ -6,8 +6,7 @@
   ],
   "plugins": [
     ["@babel/proposal-class-properties", { "loose": true }],
-    ["@babel/proposal-object-rest-spread", { "loose": true }],
-    "emotion"
+    ["@babel/proposal-object-rest-spread", { "loose": true }]
   ],
   "comments": false
 }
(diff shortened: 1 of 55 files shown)


Noviny:
Rebased to pull in changes - deleted a bit in the gatsby-config that didn't do anything, which I think the conflict was comments asking if we needed it :P

alexreardon:
Can you please also remove the `emotion-babel-plugin` so that we do not accidentally use the unsupported syntax?

pksjce:
Glad my investigation was not completely baseless!

Issue #553, opened 2018-06-18
Detect unused source files


alexreardon:
From time to time I find files in `/src` that are no longer used. It would be good to have a tool to detect and warn about these files

PR #765, opened 2018-08-27
Detect unused source files, relate to issue #553

fkhadra:
Hello,

I made a tool to parse the sources and check for unused source files. The tool is still under development but it works. The sources are available [here](https://github.com/fkhadra/trash-import). It's also published to npm. Here is the ouput for this project:
screen shot 2018-08-27 at 23 48 58

For each file I removed, I made one commit cause I wanted to be sure that it won't break anything(test suite, storybook, etc..).

The file`get-window-from-ref.js`  is referenced 2 times, I deleted the one that is not used. The code is identical:
- `src/view/window/get-window-from-ref.js` -> Not Used
- `src/view/get-window-from-ref.js` -> Used

I removed the directory src/state/publish. There is no import from `src/state/publish` or any file inside that directory.

For  the file `src/state/middleware/pending-drop.js` I had to remove the spec file under `test/unit/state/middleware/pending-drop.spec.js`.

> Edit: I tried to fix some typos


alexreardon:
Hi @fkhadra, sorry for taking so long to get onto this one. 

Love your work!

I was thinking a more sustainable path forward would be to add a test which runs your check to see if there are any unused source files. That way we could fail a build if there are unused files

fkhadra:
Hey @alexreardon, don't worry for the delay.

You have a very good idea. I'll rework my program so it can used as a prebuild tool. Thank you for the feedback.

alexreardon:
I'll close this one for now, but keen to see a prebuild tool :D

PR #767, opened 2018-08-28
Droppable placeholder

alexreardon:


alexreardon:
I am super happy with this. It has consolidated the additional placeholder logic into one place rather than leaning on two separate calculations

PR #779, opened 2018-09-02
Grouping API: request for comments!

alexreardon:
## [Read RFC](https://github.com/atlassian/react-beautiful-dnd/blob/28b979ed62cb843b4928f17e00ee8ac27e68c239/docs/rfc/grouping-api.md)

I am trying to firm up the API for #511 

Process: if you want to start a discussion then please add an inline comment to the PR to start a thread. If you have general feedback you can add a comment to the PR

### DIFF
--- a/docs/rfc/grouping-api.md
+++ b/docs/rfc/grouping-api.md
@@ -1,0 +1,53 @@
+# Grouping API
+
+We will be adding the ability to add 'grouping' support to `react-beautiful-dnd`. I am attempting to create naming for the new api that feels right. The grouping feature will enable ability to hover over a Draggable to group with it. Right now I have been calling the feature 'grouping'. However, I am not sure if even that is the right name for the feature.
+
+## API
+
+### `Draggable` > isGroupingEnabled
+
+This would allow a particular Draggable to be grouped with
+
+### `Draggable` > `DraggableStateSnapshot`
+
+```diff
+export type StateSnapshot = {|
+  isDragging: boolean,
+  draggingOver: ?DroppableId,
+-  isDropAnimating: boolean,
++  dropping: ?DroppingState,
++  groupingWith: ?DraggableId,
++  groupedOverBy: ?DraggableId,
+|};
+
++export type DroppingState = {|
++  reason: DropReason,
++  duration: number,
++|};
+```
+
+We have moved from `isDropAnimating` to `DroppingState` to give some more information about a drop. This is useful in crafting customised drop styles
+
+> I am not happy with the naming of these and I am looking for input!
+
+`groupingWith`: the id of the Draggable that is currently being dragged over. Will be provided to the dragging Draggable
+`groupedOverBy`: the id of the Draggable that is currently dragging over a groupable Draggable. Will be provided to the grouping target
+
+### `DragDropContext` > `onDragUpdate` and `onDragEnd`
+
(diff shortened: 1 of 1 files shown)


theKashey:
1. From the definition of the action you are `stacking` items, not `grouping` them. But dragging one item on anther you `stack` one on top another.
2. In some cases, items _grouped_ by some reason forms `cluster`, but usually this term is used for automated processes.
3. In some editors you can group/ungroup elements, but almost always they are not in the same position on space.

- to spread something common across group of items - use group or grouping.
- to form a stack of items, by dragging on top each other - use stack or stacking.

docs/rfc/grouping-api.md
```diff
@@ -0,0 +1,53 @@
...
+export type StateSnapshot = {|
+  isDragging: boolean,
+  draggingOver: ?DroppableId,
+-  isDropAnimating: boolean,
++  dropping: ?DroppingState,
++  groupingWith: ?DraggableId,
++  groupedOverBy: ?DraggableId,
+|};
+
++export type DroppingState = {|
```
lukebatchelor: `reason` and `duration` definitely make sense to me, I assume it's `DroppingState` that feels weird?

Not sure on a great name either. `DroppingMeta`, `DroppingData`... I'll think on it

alexreardon:
Perhaps 'Grouping' is not the right language. Other ideas:

- combining
- merging

docs/rfc/grouping-api.md
```diff
@@ -0,0 +1,55 @@
...
++  reason: DropReason,
++  duration: number,
++|};
+```
+
+We have moved from `isDropAnimating` to `DroppingState` to give some more information about a drop. This is useful in crafting customised drop styles
+
+> I am not happy with the naming of these and I am looking for input!
```
alexreardon: I am thinking of moving to the language of `combine` as I think it communicates the idea more accurately than 'grouping'

```diff
type StateSnapshot = {|
  isDragging: boolean,
  dropping: ?DroppingState,
  draggingOver: ?DroppableId,
+  combineWith: ?DraggableId,
+  combineTargetFor: ?DraggableId,
|};
```

docs/rfc/grouping-api.md
```diff
@@ -0,0 +1,55 @@
+# Grouping API
```
alexreardon: Combining API :)

docs/rfc/grouping-api.md
```diff
@@ -0,0 +1,55 @@
...
+
+We will be adding the ability to add 'grouping' support to `react-beautiful-dnd`. I am attempting to create naming for the new api that feels right. The grouping feature will enable ability to hover over a Draggable to group with it. Right now I have been calling the feature 'grouping'. However, I am not sure if even that is the right name for the feature.
+
+What the consumer actually achieves with this feature may vary. By default when dropping a item over another one, the dragging item will drop _into_ the one being hovered over
+
+## API
+
+### `Draggable` > isGroupingEnabled
```
alexreardon: Right now it is actually only: `Draggable` > ~~`isGroupingEnabled`~~ `isCombineEnabled`

Which means that it is opt in / out for the whole group. This is easier, but less flexible. I am guessing there are lists where there are some items that can be combined with and some items that cannot. 

I was thinking of keeping it simple and just having it as opt in / opt for the list.

Thoughts?

Issue #724, opened 2018-08-22
Object doesn't support property or method 'find' in IE11


wzc19920413:
## Bug or feature request?

Bug

### Expected behavior
To be able to drag and drop on IE11
### Actual behavior
App can not be opened and cause error:

SCRIPT438: Object doesn't support property or method 'find'

### Steps to reproduce
I think this is related to [#320](https://github.com/atlassian/react-beautiful-dnd/issues/320)

The console error points to the file [supported-page-visibility-event-name.js](https://github.com/atlassian/react-beautiful-dnd/blob/master/src/view/drag-handle/util/supported-page-visibility-event-name.js)

Not sure whether there are other files that would cause the error
### What version of `React` are you using?
16.2.0

### What version of `react-beautiful-dnd` are you running?
9.0.0

### What browser are you using?
IE11

alexreardon:
Related: #320 

For now you can polyfill .find

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find#Polyfill

alexreardon:
I don't think we will stop using `.find`, but we should be clearer about what polyfills are required for IE11.

alexreardon:
Sure @wuweiweiwu. At this stage we would be after a guide that states that we require `.find` to be polyfilled for ie11

tgormtx:
I've added polyfills to my app and it doesn't solve the problem. Why is filter, which is support, not being used?

alexreardon:
@tgormtx this has now been resolved in `version 10` which should be shipping soon. We will no longer require any poly filling for ie 11

#838

wuweiweiwu:
Should the README be updated?

alexreardon:
Yep! PR welcome!

PR #783, opened 2018-09-05
Updating README to include note for IE 11 polyfill

wuweiweiwu:
closes: #724 

Add note to users that `Array.prototype.find` needs to be polyfilled


alexreardon:
Thanks @wuweiweiwu !!

PR #789, opened 2018-09-11
Adding images to markdown in the website

pksjce:


website/gatsby-config.js
```diff
@@ -23,6 +23,7 @@ module.exports = {
             resolve: `gatsby-remark-images`,
             options: {
               maxWidth: 740,
+              wrapperStyle: `margin: 32px 0;`,
```
alexreardon: Can you please base this on constants / grid? 👍

pksjce: I tried `require` ing the constants into this file. But it doesnt work because this is a config file that does not go through the build aka babel.

alexreardon: fair enough

Issue #796, opened 2018-09-14
Can't use in react-frame-component


jdillick:
## Bug or feature request?

Bug

### Expected behavior
When binding react-beautiful-dnd into an iframe portal, for the functionality to continue to work inside the iframe.

### Actual behavior

You will get a number of Invariant errors, such as: "provided.innerRef has not been provided with a HTMLElement."

### Steps to reproduce

```jsx
import React from 'react';
import MyDNDComponent from './MyDNDComponent';
import Frame from 'react-frame-component';

export default (props) => {
  return (<Frame><MyDNDComponent /></Frame>);
}; 
```

### What version of `React` are you using?

^16.4.1

### What version of `react-beautiful-dnd` are you running?

^9.0.1

### What browser are you using?

Chrome latest

### Demo

Notice that when instanceof is used to check an element in the iframe by the parent window, it will fail. Duck typing these checks will bring back the functionality while retaining some manner of type safety.

```
document.write('<iframe id="testing-iframe"></iframe>')

const doc = document.getElementById('testing-iframe').contentDocument;
doc.open('text/html', 'replace');
doc.write('<div id="internal-html-element"></div>');
doc.close();
doc.body.children[0] instanceof HTMLElement
// false

doc.body.children[0] instanceof document.getElementById('testing-iframe').contentWindow.HTMLElement
// true
```

alexreardon:
Each iframe get's it's own context / objects. And instanceof just checks the prototype chain - which is why this is failing.

Some sort of duck typing could be okay. Although I think it should be hidden behind some helpers and the helpers should have the correct flow types so that the library thinks that it is a HTMLElement

alexreardon:
Portaling something into an iframe is a fairly interesting use case

alexreardon:
I would be keen to see the logic in #797 pulled into some helpers to hide this crazy!

jamesmcintyre:
Just adding that the @jdillick 's branch (https://github.com/jdillick/react-beautiful-dnd/tree/bugfix/duck-type-dom-objects) worked for me (I'm using react-frame-component).

camdagr8:
@alexreardon maybe accept @jdillick 's fix and then request for the helpers?

alexreardon:
We will look at getting something like this in soon

minhtranite:
That issue block me upgrade to from version 7 to newer.

vladkucherov:
I have the same issue

alexreardon:
[...]

PR #797, opened 2018-09-14
Duck type DOM objects, to fix usage with react-frame-component

jdillick:
#796 Probably not comprehensive, but demonstrates how duck typing could accomplish the same objective without breaking iframe portal binding.


alexreardon:
I will look at getting this in soon

vladkucherov:
any updates? @alexreardon I have the same issue

PR #910, opened 2018-11-12
Correctly expose Droppable and Draggable component types

TrySound:
Currently `Droppable` and `Draggable` components are treated as `any`
which a makes development less safer.

In this diff I with @istarkov help found a way to cover connected
components with default props without loosing type information.
A few basic flow tests will let to not remove this info accidentally.

As a result unused type property was result from some stories.

Some error suppressions are not necessary so I removed them.

### DIFF
--- a/src/view/draggable/connected-draggable.js
+++ b/src/view/draggable/connected-draggable.js
@@ -1,5 +1,6 @@
 // @flow
 import { type Position } from 'css-box-model';
+// eslint-disable-next-line
 import * as React from 'react';
 import memoizeOne from 'memoize-one';
 import { connect } from 'react-redux';
@@ -264,6 +265,7 @@ const defaultProps = ({
   disableInteractiveElementBlocking: false,
 }: DefaultProps);
 
+// eslint-disable-next-line
 /*::
 class DraggableType extends React.Component<OwnProps> {
   static defaultProps = defaultProps;
--- a/src/view/draggable/connected-draggable.js
+++ b/src/view/draggable/connected-draggable.js
@@ -1,6 +1,6 @@
 // @flow
 import { type Position } from 'css-box-model';
-import { type Node } from 'react';
+import * as React from 'react';
 import memoizeOne from 'memoize-one';
 import { connect } from 'react-redux';
 import Draggable from './draggable';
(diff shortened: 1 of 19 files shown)


src/view/draggable/connected-draggable.js
```diff
@@ -1,6 +1,7 @@
 // @flow
 import { type Position } from 'css-box-model';
-import { type Node } from 'react';
+// eslint-disable-next-line
+import * as React from 'react';
```
alexreardon: Is this needed? I have had issues with this in the past. Also, I have found just importing the type to be very clear

test/test_flow.js
```diff
@@ -0,0 +1,39 @@
+// @flow
```
alexreardon: Yum!

test/test_flow.js
```diff
@@ -0,0 +1,39 @@
...
+/* eslint-disable */
+
+import * as React from 'react';
+import { DragDropContext, Droppable, Draggable } from '../src';
+
+{
+  <>
```
alexreardon: Could you please use `React.Fragment` here? I have been using that throughout the code and it would be good to keep it consistant

test/test_flow.js
alexreardon: This is a fantastic addition

src/view/draggable/connected-draggable.js
TrySound: Which issues? You use named imports all over the code. The dangerous may be using mixed imports like `import React, { Component } from 'react';` which is used all over the project too.

src/view/draggable/connected-draggable.js
```diff
@@ -258,10 +259,23 @@ const mapDispatchToProps: DispatchProps = {
...
 };
 
+const defaultProps = ({
+  isDragDisabled: false,
+  // cannot drag interactive elements by default
+  disableInteractiveElementBlocking: false,
+}: DefaultProps);
+
+// eslint-disable-next-line
+/*::
```
alexreardon: One last thing: Can you please add a comment explaining this in the code itself?

TrySound:
@alexreardon Why closed?

alexreardon:
Mistake! I though I pressed approve 🙃

Issue #921, opened 2018-11-14
draggableId does not accept 0


flipchart:
## Bug or feature request?

Bug

### Expected behavior
No error

### Actual behavior
Console error & exception: Draggable requires a draggableId

### Steps to reproduce
Create a Draggable element with a draggableId set to 0. The props checker checks for falsey values which renders the (IMO) perfectly acceptable draggableId value of 0 as false and throws the exception

### What version of `React` are you using?
16.6.3

### What version of `react-beautiful-dnd` are you running?
10.0.1

### What browser are you using?
Chrome

alexreardon:
I will fix this tomorrow

alexreardon:
Are you using the number 0 or the string “0”?

TrySound:
@alexreardon Need a null check here https://github.com/atlassian/react-beautiful-dnd/blob/master/src/view/draggable/check-own-props.js#L12
```
invariant(props.draggableId != null, 'Draggable requires a draggableId');
```

@flipchart Could you rename the issue title to something like `draggableId does not accept 0`

flipchart:
@alexreardon using the number `0`. 

@TrySound issue renamed. I would suggest 

    invariant(props.draggableId != null && props.draggableId != undefined, 'Draggable requires a draggableId');

unless the `undefined` check is done elsewhere?

flipchart:
Aaah, thanks. I'm so used to triple equals (due to TSLint) that I'm not aware of the coercion rules with double equals

alexreardon:
I can loosen the check, but keep in mind:

- we might be doing truthy checks elsewhere, so using 0 might not work anyway
- we call out that `draggableId` needs to be a string. If you want you can just use `"0"` as the `draggableId`
- even if i get it working today **there will be regressions** because it is a violation of our public types and so we won't be testing against it..

the more i think about it, the more i think maybe we should not account for this. Thoughts @lukebatchelor @TrySound ?

TrySound:
@alexreardon Flow has a lint rule to safely handle such case.
https://flow.org/en/docs/linting/rule-reference/#toc-sketchy-null

TrySound:
However it won't check this specific case since type is not nullable.

alexreardon:
Based on [discussion](https://github.com/atlassian/react-beautiful-dnd/pull/925#issuecomment-438862750) in #925 I am leaning towards declining this PR. 

We only accept officially accept `string` ids at this stage.

[...]

Issue #498, opened 2018-05-10
Any way to pass state/callback to onDragEnd from Draggable (and Droppable)?


spatialvlad:
I really like the great library that you guys have created. Our team is currently considering the option to start using react-beautiful-dnd instead of react-sortable-hoc we currently use.

However, it looks like there is a blocker for adopting your library for us due to inability to pass state/callback from Draggable into the onDragEnd of the DragDropContext. I hope I just missed the option in the documentation, but I did make a thorough search before opening the issue.

Your documentation recommends wrapping whole application with a DragDropContenxt. For most real-life production applications it means that DragDropContext will be multiple levels of hierarchy away form the Draggables. Even if multiple DragDropContexts are used, it is still very likely that the DragDropContext (probably with Droppable) will be not in the same file/component as the Draggables.

It is definitely not a big issue to create something like a top-level reducer inside of the onDragEnd that first filters results based on type, then applies corresponding logic, getting needed data from the Redux store. And this solution will work in some cases.

However, there are several drawbacks and even blockers with this approach:

It is extra work to get the Redux store data and process it in the corresponding onDragDrop reducer. This work was already probably conducted at the component level - Draggable or Droppable. And simply passing it would be much easier than trying to do it again.
Much bigger problem is that not all Component state lives in Redux store. This is a quite common and even recommened (e.g. https://github.com/reactjs/redux/issues/1287 ) behavior. This means that onDragEnd currently simply does not have a way to get needed state, unless substantial refactoring of existing code is done to put all needed state in a Redux store (and even then it might be not convenient).

Other libraries do not seem to have same limitation, as both offer a callback on the Draggable level to run after dragging ends.

E.g. React DnD offers endDrag for the DragSource. React-sortable-hoc has onSortEnd for the SortableComponent.

Your library looks superior to the above mentioned libraries in many aspects, but current limitation of not being able to define/pass a onDragEnd callback or at least some state from the Draggable is unfortunately greatly undermining its usability, especially if we are talking not about greenfield projects, but ones already developed and have lots of (presentational) state on the components level.

I hope I just missed some piece of the documentation and you can direct me where to have a look.

I did find something that seems related (but still does not seem to answer my question):
https://github.com/atlassian/react-beautiful-dnd/issues/120
https://github.com/atlassian/react-beautiful-dnd/issues/466

It would be really great if there actually was some way to pass a callback from Draggable. I would appreciate if you could help me find out how to do this.

Thanks.

[...]

PR #925, opened 2018-11-14
Enable sketchy null check

TrySound:
Ref https://github.com/atlassian/react-beautiful-dnd/issues/921

This lint rule allows to not possible errors with falsy values.


TrySound:
This PR does NOT fix this issue https://github.com/atlassian/react-beautiful-dnd/issues/921

alexreardon:
This is a great fix. Thanks @TrySound. I am thinking more critically about whether we *should* do this 🤔

alexreardon:
This has reasonably big code impact to support a use case that our flowtypes say we do not support. I wonder if this will lead us down a rabbithole of 'supporting' things that we say we don't. I think if people are going to ignore the flow types, well, is that on them?

🤔

TrySound:
Well, this may fail with empty string too. Sadly there's not sketchy-string rule. I think some cases should be covered with flow, some like this with tests. There are a lot of users without flow.

alexreardon:
If we wanted to be really strict, we could do this when we do the prop check:

```js
invariant(typeof draggableId === 'string', 'draggableId must be a string');
invariant(draggableId.length, 'draggableId cannot be an empty string');
```
This would force `draggableId` to be a string.

However, we are a bit loose with the check right now as people are using `draggableId` to pass data as we do not have a `payload` prop yet #498. Currently it is being used to pass functions in some cases 😲.

It feels like an alternate path forward to this PR is to create the `payload` prop and be stricter in our checks about the type and shape of `draggableId`

Thoughts @TrySound?

I am also worried that this check might encourage an incorrect use of booleans:

> false == null
false 
(thanks @lukebatchelor)

TrySound:
I think it's okay to compare with `true` or with `false` depending on default. Did I made a mistake somewhere?

Yeah, typeof will fit better in this case.

alexreardon:
No you did not make a mistake. Your PR is great! But i wonder if forcing people internally to check against null would accidentally lead people to check booleans against null

TrySound:
Well, logical problems should be caught by tests. Mistake may happen without null check too. This at least enforces to cover usually hidden problems.

Issue #794, opened 2018-09-12
I need specific behavior of 3 lists


drebits:
## Bug or feature request?

Bug, I guess.

### Expected behavior
Hi, I'm following the [egghead tutorials](https://egghead.io/lessons/react-conditionally-allow-movement-using-react-beautiful-dnd-draggable-and-droppable-props) and I'm trying to get the next behavior:
I have 3 lists, and I need to drag and drop items from the last list to any other list, and between the first 2 lists too, but not from the first 2 lists to the third one.
But not works, even as in the tutorial.

### Actual behavior

I get this error: **Trying to set droppable isEnabled to true but it is already true.**

### Steps to reproduce

I have this [code repo](https://github.com/drebits/beauty-dnd) and I do things like in this [lesson](https://egghead.io/lessons/react-conditionally-allow-movement-using-react-beautiful-dnd-draggable-and-droppable-props).

### What version of `React` are you using?
```
"react": "^16.5.0",
"react-dom": "^16.5.0",
```
### What version of `react-beautiful-dnd` are you running?
```
"react-beautiful-dnd": "^9.0.2",
```
### What browser are you using?

google-chrome@69.0.3497.92

### Demo

I have this base code [https://github.com/drebits/beauty-dnd](https://github.com/drebits/beauty-dnd), and I need the behavior described above. Thanks any help.

alexreardon:
Hi there! Thanks for raising this. 

Can you please create a stand-alone example on codesandbox? Here is a boilerplate you can start from https://codesandbox.io/s/k260nyxq9v

drebits:
Hi @alexreardon, thanks for your answer, here I have the example https://codesandbox.io/s/z2jlm9x243
I need the last list (droppable) be disabled when you try to drop from others lists. But when the dragging starts returns the error.

alexreardon:
Okay, I know the issue: you are passing in isDragDisabled as `null`. This is an invalid value. It either needs to be a `boolean` or not provided (`undefined`) If you passed in `undefined` you would get our default value of `false`.

`flow` or `typescript` would have picked this up.

I am not sure if it is worth adding a runtime check for this given that it is an invalid value...

alexreardon:
Passing in `null` is a violation of our types. I am not sure how far to go with run time checking. Please either pass in `undefined` or cast your value to a boolean `Boolean(myValue)`

PR #945, opened 2018-11-21
More dev only prop checks

alexreardon:


alexreardon:
Would have prevented #794

Issue #954, opened 2018-11-24
Broken Link at react-beautiful-dnd/docs/patterns/multi-drag.md


modemmute:
There's a broken link in [react-beautiful-dnd/docs/patterns/multi-drag.md](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/patterns/multi-drag.md) in the second paragraph.

"We have created a [reference application](https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/patterns/react-beautiful-dnd.netlify.com/iframe.html?selectedKind=Multi%20drag&selectedStory=pattern&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel) ([source](https://github.com/atlassian/react-beautiful-dnd/blob/master/stories/9-multi-drag-story.js)) which implements this multi drag pattern" -- the 'reference application' link is a 404 and the 'source' link doesn't seem to link to any relevant code.

alexreardon:
This has now been fixed by #955. Thanks @modemmute

Frankcarvajal:
The issue has not been fixed, and should be reopened. 
I created a new issue #1118 

Cheers!

PR #955, opened 2018-11-25
Fix dead doc links that were not root-relative

Blasz:
I updated the non root-relative ones in README.md as well even though they weren't broken just so they don't break if they're ever extracted out into their own document.


alexreardon:
Thanks @Blasz !!!

Frankcarvajal:
@alexreardon I think that these 2 commits did not resolve #954

PR #960, opened 2018-11-27
adding test for client hydration

alexreardon:


alexreardon:
Inspired by @petegleeson

PR #966, opened 2018-11-28
fix: droppable-get-dimension withScroll

woohling:
We're forking your amazing repo and need to add some features based on our product logic, and then I found a potential bug, which I spent a lot of time debuging.

The default container supporting horizontal scroll is the window, which would not trigger any bugs currently. But I think it would be stricter to code it this way. Pls take a look and see if I'm doing this correctly.

### DIFF
--- a/src/view/droppable-dimension-publisher/get-dimension.js
+++ b/src/view/droppable-dimension-publisher/get-dimension.js
@@ -117,7 +117,7 @@ export default ({
 
     return {
       client: frameClient,
-      page: withScroll(frameClient),
+      page: withScroll(frameClient, windowScroll),
       scroll: getScroll(closestScrollable),
       scrollSize,
       shouldClipSubject,


woohling:
@alexreardon kindly pls take a look at the pull request

alexreardon:
Away this week. Will look next week

src/view/droppable-dimension-publisher/get-dimension.js
```diff
@@ -117,7 +117,7 @@ export default ({
 
     return {
       client: frameClient,
-      page: withScroll(frameClient),
```
alexreardon: `withScroll` will apply the page scroll

alexreardon: You are right that your change is a bit stricter in that it does not read the page scroll directly, but just supplies it..

Thoughts?

alexreardon: I also like that this approach does not touch the DOM again and just uses the value that has been passed in

PR #967, opened 2018-11-29
Dynamically adding area to foreign list when dragging over

woohling:
#77 
I'm fixing this issue by adding extra area when dragging over foreign list. I hope I'm fixing it right.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 339807,
-    "minified": 130844,
-    "gzipped": 38870
+    "bundled": 341203,
+    "minified": 131465,
+    "gzipped": 39004
   },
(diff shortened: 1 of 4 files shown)


alexreardon:
This increase is already achieved but the `withPlaceholder` function 👍

alexreardon:
Thanks for being the change you seek :)

woohling:
I know with placeholder can increase the droppable space.  My update is for when the droppable height is fixed and the last one on the list is quite hard to target at. 

For example, when I drag an item over the list on item 12 and want to put the item behind it, the hit rate is quite low, cause the active area is the height of the list before dragging start. What I'm proposed is to increase the active area when dragging over the list, thus I can easily put the dragging item on the last spot of thelist.
xnip2018-12-03_13-29-49

woohling:
@alexreardon I'm closing this Pull Request cause I found that it will cause bugs in some scenario. More scenarios need to be taken into control, I'm still too careless

PR #976, opened 2018-12-03
Auto scroll grab near edge

### LINKED ISSUE #349: Avoid auto scrolling when picking up a Draggable near a container edge
Currently if you pick up a `Draggable` near the edge of a scroll container or the window and auto scroll will start immediately. I think we could improve this.

Idea: only allow auto scrolling after the user has moved in the direction of the edge we are trying to scroll.

alexreardon:
Fixes #349


alexreardon:
I hope to merge this on Monday. Friday is a bad time for releases

Issue #1007, opened 2018-12-12
flow crashing on circleci


alexreardon:
Builds have been increasingly running into this issue:

```sh
$ flow check --max-warnings=0
Worker interrupted with signal: sigkill
Subprocess(364): fail 2Unhandled exception: Worker_failed (process_id = 364): (Worker_quit WEXITED 2)
Raised at file "src/core/lwt.ml", line 3008, characters 20-29
Called from file "src/unix/lwt_main.ml", line 42, characters 8-18
Called from file "src/commands/checkCommands.ml", line 135, characters 27-86
Called from file "src/commands/commandUtils.ml", line 13, characters 4-32
```

I have sshed into the box and run

`yarn flow` and it cannot spin up the server. It tries 7 times then exits

alexreardon:
Done ^

alexreardon:
It looked like that worked @TrySound. Super strange. What do you think would be the best way to only use max_workers for ci?

alexreardon:
I will re-run the build of #1008 to see if it works consistantly

TrySound:
Not sure, but some time ago i saw some custom config specifying. Will take a look tomorrow.

alexreardon:
Thanks @TrySound

alexreardon:
You're a hero! 💪

TrySound:
The problem is circle ci. O had the same problem with jest workers

alexreardon:
I have merged that change for now. Ideally we only have this setting on for ci, but running local seems to be fine with max_workers=1. 

I have merged for now because it fixes our builds

jbrown215:
This looks resolved to me-- sorry I didn't get to this earlier.

PR #1008, opened 2018-12-12
setting max workers

alexreardon:
#1007


alexreardon:
This looks good to me. I will run it a few more times to see if we get any issues

alexreardon:
I think this is a similar build issue to what we where getting previously with jest

Issue #1021, opened 2018-12-18
Draggables not working for heavy handed touch


markbrown4:
## Bug

A simple vertical list with sortable div's does not work reliably on touch devices, (tested on iPhone6).
Only when you touch lightly and hold does it work, if you're too heavy handed you can't use the Draggables at all.

The demo example shows this behaviour:
https://codesandbox.io/s/k260nyxq9v

### Expected behavior

For simple lists without links you should be able to drag on the handle reliably every time, regardless of pressure.

I believe this is related to the force touch support that was added for links, when there's no links and no need for force touch I'm hoping this can improved, otherwise we're going to need to fallback to up / down buttons on mobile, it doesn't work enough of the time and makes for a worse experience.

**Versions:**
`react`: 16.6.3
`react-beautiful-dnd`: 10.0.3

**Browsers:**
Tested on ios 12.1.1:

- Mobile Safari
- Chrome
- Firefox

alexreardon:
Here is the relevant code:

```js
// Need to opt out of dragging if the user is a force press
    // Only for webkit which has decided to introduce its own custom way of doing things
    // https://developer.apple.com/library/content/documentation/AppleApplications/Conceptual/SafariJSProgTopics/RespondingtoForceTouchEventsfromJavaScript.html
    {
      eventName: 'touchforcechange',
      fn: (event: TouchEvent) => {
        // A force push action will no longer fire after a touchmove
        if (state.hasMoved) {
          // This is being super safe. While this situation should not occur we
          // are still expressing that we want to opt out of force pressing
          event.preventDefault();
          return;
        }

        // A drag could be pending or has already started but no movement has occurred

        const touch: TouchWithForce = (event.touches[0]: any);

        if (touch.force >= forcePressThreshold) {
          // this is an indirect cancel so we do not preventDefault
          // we also want to allow the force press to occur
          cancel();
        }
      },
    },
```

Right now we cancel a drag if we detect a force press. 

Here is the relevant section from the README:

> ### Understanding intention: tap, force press, scroll and drag
> 
> When a user presses their finger (or other input) on a `Draggable` we are not sure if they where intending to _tap_, _force press_, _scroll the container_ or _drag_. **As much as possible `react-beautiful-dnd` aims to ensure that a users default interaction experience remains unaffected**.
> To see more indepth information about how we impact standard browser events see our [how we use DOM events guide](/docs/guides/how-we-use-dom-events.md)

[...]

PR #1022, opened 2018-12-18
Allow force touch behaviour to be configured via props

alexreardon:
From #1021

### DIFF
--- a/src/view/drag-handle/drag-handle.jsx
+++ b/src/view/drag-handle/drag-handle.jsx
@@ -56,6 +56,7 @@ export default class DragHandle extends Component<Props> {
       getDraggableRef: this.props.getDraggableRef,
       getWindow,
       canStartCapturing: this.canStartCapturing,
+      canCancelDragWithForceTouch: this.props.canCancelDragWithForceTouch,
     };
 
     this.mouseSensor = createMouseSensor(args);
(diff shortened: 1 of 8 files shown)


src/view/drag-handle/sensor/create-touch-sensor.js
```diff
@@ -357,7 +358,7 @@ export default ({
 
         const touch: TouchWithForce = (event.touches[0]: any);
 
-        if (touch.force >= forcePressThreshold) {
+        if (touch.force >= forcePressThreshold && canCancelDragWithForceTouch) {
```
alexreardon:
```suggestion
if (canCancelDragWithForceTouch && touch.force >= forcePressThreshold) {
```

src/view/drag-handle/drag-handle.jsx
```diff
@@ -45,6 +45,10 @@ export default class DragHandle extends Component<Props> {
     [canLiftContextKey]: PropTypes.func.isRequired,
   };
 
+  static defaultProps = {
+    canCancelDragWithForceTouch: true,
```
alexreardon: this should be a standard prop for a drag handle an an optional prop on the Draggable (with a defaultProp)

alexreardon:
Let’s hold off on this one for a bit. I want to think it through a bit

Issue #1010, opened 2018-12-13
Christmas maintenance 🎄(mid dec - mid jan)


alexreardon:
### Merry Christmas everyone! 🎄🎅🎉

I will be away for a few weeks for a Christmas and end of year break. 

@treshugart will be keeping the lights on during my absence. Thanks @treshugart !!!

I plan on returning in mid-January

Thanks for a great 2018!!

alexreardon:
While away I have updated my watch mode to ignore 👌

alexreardon:
I will take a look at the issues linked to this one soon

PR #1037, opened 2019-01-01
Added CSP nonce

Zweder:
I added a nonce property to DragDropContext which is added to the injected style tags to make this project work with strict ContentSecurityProtection settings


treshugart:
Thanks, @Zweder. Alex will have a look at this when he's back. See #1010 for more details on his absence.

alexreardon:
This looks good to me. I'll have a play. @Zweder can you confirm that this fixes #1050?

alexreardon:
For now let's see if this resolves the issue. After that we will need to do some productisation:

- commenting code to explain why nonce is there
- adding to docs to explain nonce prop

alexreardon:
There are a number of CSS in JS solutions that add css to a page without requiring this sort of thing. I think it would be worth looking into how they get around this issue

alexreardon:
Let's pick this up: https://github.com/atlassian/react-beautiful-dnd/issues/1050#issuecomment-485605440

funktr0n:
Hey guys, distant observer here but I just wanted to see where things were with this. Is this PR the only way forward for nonce support or are there other alternatives now?

Zweder:
Right back on it. A cypress test will be up next

alexreardon:
Heck yes!

browser-test-harness.js
```diff
@@ -9,29 +9,52 @@ const storybook = childProcess.spawn(process.execPath, [
...
+  path.join('test', 'standalone', 'start.sh'),
+  '9003',
+]);
+
 process.on('exit', () => {
   storybook.kill();
+  standalone.kill();
 });
 
 waitPort({
```
alexreardon: can you explain this change?

docs/api/drag-drop-context.md
```diff
@@ -1,6 +1,6 @@
...
+In order to use drag and drop, you need to have the part of your `React` tree that you want to be able to use drag and drop in wrapped in a `<DragDropContext />`. It is advised to just wrap your entire application in a `<DragDropContext />`. Having nested `<DragDropContext />`'s is _not_ supported. You will be able to achieve your desired conditional dragging and dropping using the props of `<Droppable />` and `<Draggable />`. You can think of `<DragDropContext />` as having a similar purpose to the [react-redux Provider component](https://react-redux.js.org/api/provider). A content-security-protection nonce attribute is added to the injected style tags if provided.
```
alexreardon: Cool

package.json
```diff
@@ -130,6 +130,7 @@
     "stylelint-config-styled-components": "^0.1.1",
     "stylelint-processor-styled-components": "^1.8.0",
     "testcafe-reporter-xunit": "^2.1.0",
+    "uuid": "^3.3.2",
```
alexreardon: Can we just use a local sequence generator to avoid another dependency?

Eg:
```js
let count = 0;
function getId(): string {
  return `id-${count++}`;
}
```

alexreardon:
Thanks for your hard work on this one. I think it is close to being good to go. I just had a few minor thoughts

cypress/integration/reorder-lists.spec.js
```diff
@@ -2,11 +2,12 @@
 import * as keyCodes from '../../src/view/key-codes';
 import { timings } from '../../src/animation';
 
-beforeEach(() => {
-  cy.visit('/iframe.html?id=board--simple');
-});
-
 it('should reorder lists', () => {
+  cy.visit('/iframe.html?id=board--simple', {
+    headers: {
```
alexreardon: love this

alexreardon:
It is fantastic that we are actually testing this in a browser. Well done

test/standalone/server.js
```diff
@@ -0,0 +1,62 @@
+// @flow
+import express from 'express';
+import React from 'react';
+import { renderToString } from 'react-dom/server';
+import uuidv4 from 'uuid/v4';
+import { resolve } from 'path';
+import Sample from './app';
+import { resetServerContext } from '../../src';
+
+const getNonce = () => Buffer.from(uuidv4()).toString('base64');
```
alexreardon: see other comment about just using a simple alternative for this one rather than adding a new dependency

test/standalone/webpack.config.js
```diff
@@ -0,0 +1,81 @@
...
+              // limit: false,
+            },
+          },
+        ],
+      },
+      {
+        test: /jsx?$/,
+        // type: "javascript/auto",
+        exclude: [/node_modules/],
+        use: [
```
alexreardon: can we defer to our standard babel rc file for this?

test/standalone/main.js
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable id-length */
+// @flow
+
+const webpack = require('webpack');
+const requireFromString = require('require-from-string');
+const MemoryFS = require('memory-fs');
+const path = require('path');
+const config = require('./webpack.config');
+
+const fs = new MemoryFS();
```
alexreardon: why is this used?

test/standalone/main.js
```diff
@@ -0,0 +1,23 @@
...
+const compiler = webpack(config);
+
+// $ExpectError
+compiler.outputFileSystem = fs;
```
alexreardon: can you explain this one to me? I am not a webpack ninja

NFabrizio:
Just a general question. Would it be preferable to either rely on a meta tag instead of a passed prop or allow both? 

I ask this since there are a few other packages that pull in react-beautiful-dnd that may or may not allow the passing of a nonce from the consumer to the react-beautiful-dnd component. 

A decent example of this is JSS which uses a meta tag named "csp-nonce". Here is how they are implementing their use of that meta tag. 
https://github.com/cssinjs/jss/blob/0af646d31962518f9603493825d43f3410ff6c14/packages/jss/src/DomRenderer.js#L255

Their implementation allows downstream consumers to pass a nonce to their components effectively without having to rely on the middleman to pass the nonce through.

package.json
Zweder: I should be a cryptographically save random number but to prove a test i will change it

browser-test-harness.js
Zweder: standalone is a mini webpack server and client instance, the nonce has to be generated at the server

package.json
Zweder: This looks better i think

`let count = 0;
function getNonce(): string {
  return `ThisShouldBeACryptographicallySecurePseudoRandomNumber-${count++}`;
}`

test/standalone/server.js
Zweder: Fixed

test/standalone/main.js
Zweder: This is the setup for the standalone webpack server and client, which run from memory saves a lot of pollution of output files

test/standalone/main.js
Zweder: I've never managed to get Flow happy about specifiing a outputFileSystem, but i dont like to cleanup everything webpack outputs

test/standalone/webpack.config.js
Zweder: Yes, cleaned up

alexreardon:
That seems really smart @NFabrizio. What do you think @Zweder ?

Zweder:
You are right @NFabrizio, actually thats the whole idea behind this PR but in a different way. Accessing the dom is not efficient thats why you schould provide the nonce with the help of React.Context which in turn gets the nonce from a meta tag or something else. That means that you only access the dom once in the lifetime of the app, and you can reuse that context provider for every third party component that needs a nonce.

In src/test/standalone/server.js i do add a meta tag csp-nonce and in src/test/standalone/client.js it picks up the meta tag and provides the nonce to the DragDropContext in src/test/standalone/app.jsx without React.Context for simplicity

Zweder:
@alexreardon It would be nice if you could find some time to review my commits

alexreardon:
This looks good to me. Can we please target the `dev` branch? (`12.0`). It would be great to get this in!

Zweder:
@alexreardon On it, i've got to fix the tests, the rest was a easy merge

Zweder:
@alexreardon this PR is now in sync with the dev branch. Some tests are failing but are not related to this PR.

alexreardon:
I'll try to take a look at this soon 👍

docs/guides/common-setup-issues.md
```diff
@@ -8,11 +8,11 @@ For detectable setup issues we try to log information in the `console` for `deve
 
 ## React version
 
-Please ensure that you meet our peer dependency version of `React`. Your React version needs to be greater than or equal to `16.3.1`.
```
alexreardon: Cheers!

alexreardon:
This is looking really good! One more thing: can you create a `content-security-policy` guide? Explaining what the problem is, and how you use a `nonce` to get going

alexreardon:
#1317

alexreardon:
If you do not have time, I can try to put something together

Zweder:
I can make time upcoming weekend, it's not something I can do in 5min

Op do 26 sep. 2019 03:16 schreef Alex Reardon <notifications@github.com>:

Zweder:
I've added content-security-policy it's insired on the JSS CSP docs but it shows how to use it and explains what it's for. Have a look

alexreardon:
Fantastic work on this!

alexreardon:
👏 This will be going out in `12.0` #1487

Issue #1036, opened 2018-12-31
Touch doesn't work in Microsoft Edge


lilymatcha:
### Expected behavior

You long-press to select in order to initiate the drag.

### Actual behavior

When you long-press, it simply selects and does not drag.

### Steps to reproduce

Open Microsoft Edge, go to any of the examples, and attempt to use touch to drag.

### What version of `React` are you using?

16.6.3

### What version of `react-beautiful-dnd` are you running?

10.0.3

### What browser are you using?

Microsoft Edge (v. HTML 17.17134)

### Demo

On [Youtube](https://www.youtube.com/watch?v=oLTnS_j6vGU&feature=youtu.be).

Weffe:
What device is that demo running on? Microsoft Surface Pro?

lilymatcha:
@Weffe Microsoft Surface Laptop

treshugart:
Alex will be back mid-Jan. See #1010 for more details.

alexreardon:
I tested this and it seemed to work for me.. Strange. Should continue to investigate!

I will try to get access to appropriate hardware

lilymatcha:
@alexreardon What hardware are you testing on? I think it shouldn't be a hardware (i.e. lack of drivers, etc.) issue, because the library works fine for me on Chrome, just not on Edge. Furthermore, adding the pointer sensor (#1045) fixed the issue for me, so I'm curious what's different that could be allowing it to work on your end.

alexreardon:
I mentioned this in #1045:

> Edge is moving to Chromium and there are already beta Edge builds. My expectation is that Edge moving to Chromium will fix #1036. Please let me know if that is not the case and I am more than happy to revisit this.
> I want to stress how thankful I am @lilymatcha for your input. Let me know if there are other parts of this library you are keen to get involved in

[Side thought about edge](https://twitter.com/alexandereardon/status/1117583118423056384)

PR #1045, opened 2019-01-07
Add pointer sensor so that touch works in Edge

lilymatcha:
### What change is being made?

I added a new sensor, `PointerSensor`, which is based on `PointerEvent`s as opposed to `MouseEvent` or `TouchEvent`. This sensor gets used as a last resort after the other sensors.

### Why is this change being made?

Previously, touch did not work in Edge (#1036). This was because while Chrome makes use of `MouseEvent` and `TouchEvent`, Edge makes use of `PointerEvent`. `PointerEvent` encapsulates mouse, touch, and pen events, so by adding a `PointerEvent`-based sensor, we're able to add touch compatibility to Edge. It likely also adds pen compatibility, but I do not have a pen device to test with. 😐

### How was this change tested?

I tested this change by:

1) Running `yarn build:dist` from my local fork of `react-beautiful-dnd`.
2) Copying over the compiled `react-beautiful-dnd.js` file to my local project's `node_modules/react-beautiful-dnd/dist/`. The code for this project can be found at at [lilymatcha/rearrangeable-layout](https://github.com/lilymatcha/rearrangeable-layout).
3) Running `npm run start` from my local testing project.
4) Attempting drag-and-drop using both mouse and touch on the latest versions of Chrome, Edge, and Firefox, and ensuring that the behavior is as expected.

### Specific requests / things to consider

- This is my first time contributing to an open-source project, so if there's any steps or anything I'm missing, or if you have any feedback, please let me know! Please feel free to comment on the code if you'd like, I would love to receive feedback and learn so I can improve.
- I wasn't able to get the entire project to build--couldn't get the `.esm.js` files to build--even from my initial fork, before I made any changes, so I've just been testing using the plain `.js` compiled file. Because this problem existed before I made any changes and didn't change throughout my testing process, I don't think my changes are the source of this error. I can provide more information on the error I ran into if needed.
- What's the correct process for updating `.size-snapshot.json` such that an updated size (from the addition of new code) gets accepted into the new build? This is what's causing the "bundle-test" CircleCI test to fail.
- The "test" CircleCI test is failing due to what seems like a browser connection problem, but I don't have permissions to rerun the test. Would it be possible to trigger a re-run of the test to ensure that it's just a browser connection issue and not a code failure?

Note: tests are forthcoming! I'm just having trouble getting the test suite to run, but once I get that figured out, I'm going to mimic the tests written for the mouse sensor.

I realize @alexreardon is out on vacation (happy holidays!) and so won't be able to take a look at this until mid-January, which is totally fine. In the meantime, if anyone wanted to take a look and leave a comment, I'd welcome the feedback! Thank you :)


jdrage:
Great! As we discussed in person you're going to revert the changes to yarn.lock which is likely causing the checks to fail.

src/view/drag-handle/drag-handle.jsx
```diff
@@ -19,19 +20,28 @@ import shouldAllowDraggingFromTarget from './util/should-allow-dragging-from-tar
...
   event.preventDefault();
 };
 
-type Sensor = MouseSensor | KeyboardSensor | TouchSensor;
+type Sensor = MouseSensor | KeyboardSensor | TouchSensor | PointerSensor;
+
+let IS_TOUCH_COMPATIBLE: boolean = false;
+
+window.addEventListener('touchstart', function onFirstTouch() {
+  IS_TOUCH_COMPATIBLE = true;
```
jdrage: Seems like on a touch compatible system this must happen before any PointerEvent happens. Maybe set something on the first PointerEvent and add an invariant here that checks that has not been set?

src/view/drag-handle/sensor/create-mouse-sensor.js
```diff
@@ -324,4 +324,4 @@ export default ({
   };
 
   return sensor;
-};
+};
```
jdrage: Whitespace only change, revert this file

src/view/drag-handle/drag-handle.jsx
```diff
@@ -192,31 +198,56 @@ export default class DragHandle extends Component<Props> {
...
+    if (this.keyboardSensor.isCapturing() || this.mouseSensor.isCapturing() || this.pointerSensor.isCapturing()) {
       return;
     }
-
     this.touchSensor.onTouchStart(event);
   };
 
+  onPointerDown = (event: PointerEvent) => {
+
+    if (!IS_TOUCH_COMPATIBLE) {
```
jdrage: Good optimization here.

jdrage: Although in a non-touch-compatible environment you will end up adding this each time. Could store a separate Boolean for having added the listener.

src/view/drag-handle/sensor/create-pointer-sensor.js
```diff
@@ -0,0 +1,220 @@
...
+          schedule.move(point);
+          return;
+        }
+
+        // There should be a pending drag at this point
+
+        if (!state.pending) {
+          // this should be an impossible state
+          // we cannot use kill directly as it checks if there is a pending drag
+          stopPendingDrag();
```
jdrage: Maybe event.preventDefault() here for safety?

alexreardon:
@lilymatcha I am extremely thankful for the work you have done on this pull request.

Edge is moving to Chromium and there are already [beta Edge builds](https://blogs.windows.com/windowsexperience/2019/04/08/microsoft-edge-preview-builds-the-next-step-in-our-oss-journey/). My expectation is that Edge moving to Chromium will fix #1036. Please let me know if that is not the case and I am more than happy to revisit this.

I want to stress how thankful I am @lilymatcha for your input. Let me know if there are other parts of this library you are keen to get involved in

PR #1064, opened 2019-01-22
Fix closing tag in docs

hannesvdvreken:
closed with div instead of Person


alexreardon:
Thanks @hannesvdvreken !!

PR #1067, opened 2019-01-23
more robust typeof element checks

alexreardon:
Closes #797 #796


alexreardon:
I figured out a way to do this without relying on duck typing 👍👍👍

Maintaining the same type safely while allowing iframes

wow-jake

alexreardon:
cc @jdillick @vladkucherov @SenyaMur

SenyaMur:
Thank you! When you will publish this solution?

alexreardon:
Soon 🤞

PR #1078, opened 2019-01-29
animating placeholders (old attempt)

alexreardon:


alexreardon:
Now being done in #1096

PR #1081, opened 2019-01-30
adding iframe pop decorator

alexreardon:
screen shot 2019-01-30 at 11 19 12 am


alexreardon:
I have been doing this manually for too long

PR #1086, opened 2019-01-31
animated placeholders (attempt 2)

### LINKED ISSUE #77: Animated placeholders when transitioning between lists

alexreardon:
Closes #77


worsnupd:
Any ETA on this feature? This would be extremely useful to have!

alexreardon:
See #1096

PR #1128, opened 2019-02-28
Emotion again

alexreardon:


varHarrie:
Why `Emotion`? Just be curious.

alexreardon:
> Why `Emotion`? Just be curious.

There is almost nothing between them. We are thinking of consolidating internally on emotion and I wanted to give it a go for a while.

Longer answer: 

- performance: similar
- kbs: emotion is a bit smaller (not super important for our doc site)
- emotion has a more flexible api which can help us achieve behaviours that are harder with styled-components
- emotion has a cleaner SSR story
- emotion has a styled.div syntax so you can use it exactly the same and you would styled-components. This also allows for a simpler rollback path.

Given it was so easy to change, we might change back in the future too

alexreardon:
moving to emotion again

varHarrie:
Thank for your clear explanation. I am a heavy user for `styled-components` and a defector of `emotion`. I had forgotten the reason since they are too similar.

PR #1187, opened 2019-03-21
Draggable refactor

### LINKED ISSUE #1178: draggableSnapshot.draggingOver when dropping over nothing
## Bug
The behavior of `draggableSnapshot.draggingOver` seems to have changed in `10.1.0`.  I updated from `10.0.3` and found that when dropping an item outside of a droppable area,
`draggableSnapshot.draggingOver` now reports the source droppable id, instead of `null`.  This occurs only while `draggableSnapshot.isDropAnimating` is `true` - the value is `null` before and after the drop.  

To me this makes it impossible to discern a drop outside of the source container vs. a drop inside of it.  I am using a combination of `draggingOver` and `isDropAnimating` to apply a class to the draggable item while it is animating, but only when dropped outside of it's source container.  It also seems inconsistent as the value is `null` before and after the drop, while I am outside of valid droppable containers.  It still correctly reports the ids of droppable containers as I am dragging/dropping over them.

Thanks in advance, stellar library overall 👍 

### Expected behavior
`draggingOver === null` <--- this is the behavior in `10.0.3`

### Actual behavior
`draggingOver === "[droppable_id_of_source_container]"` <--- this is the behavior in `10.1.0`

`16.8.0`

`10.1.0`

alexreardon:
Fixes #1178


src/view/draggable/connected-draggable.js
```diff
@@ -32,54 +32,83 @@ import type {
...
+const getCombineWithFromImpact = (impact: DragImpact): ?DraggableId => {
+  return impact.merge ? impact.merge.combine.draggableId : null;
 };
 
 // Returning a function to ensure each
 // Draggable gets its own selector
 export const makeMapStateToProps = (): Selector => {
-  const memoizedOffset = memoizeOne(
-    (x: number, y: number): Position => ({ x, y }),
+  const getDraggingSnapshot = memoizeOne(
```
alexreardon: moved snapshot creating to connected component

src/view/draggable/connected-draggable.js
```diff
@@ -178,29 +224,40 @@ export const makeMapStateToProps = (): Selector => {
 
       const dimension: DraggableDimension =
         state.dimensions.draggables[ownProps.draggableId];
-      const draggingOver: ?DroppableId = whatIsDraggedOver(completed.impact);
-      const combineWith: ?DraggableId = getCombineWith(completed.impact);
+
+      const result: DropResult = completed.result;
+      const mode: MovementMode = result.mode;
+      // these need to be pulled from the result as they can be different to the final impact
+      const draggingOver: ?DroppableId = getDraggingOverFromResult(result);
```
alexreardon: fixes #1178

PR #1188, opened 2019-03-21
Refactoring droppable

alexreardon:
The second part of #1178


alexreardon:
I have noticed a minor visual issue with this that I will look at

Issue #871, opened 2018-10-26
RFD: Moving to a hook api 🎣


alexreardon:
[`React hooks`](https://reactjs.org/docs/hooks-intro.html) have recently been announced.

Would we benefit from moving from a render props api to a hook based api for `react-beautiful-dnd`? If so, what would it even look like?

I have not thought very much about this, but I wanted to open it for discussion

YurkaninRyan:
Oh wow this is an interesting question.  I think a hook based api would have to come along with a render prop based api.  Otherwise, you wouldn't be able to consume the hooks in non-function components.

I'm not 100% sure what a `useDraggable` or `useDroppable` would even look like.

I would assume that it would use `useEffect` and `useContext` to do its work?

YurkaninRyan:
Yup, and if we were sticking with the way that `useState` works it would probably actually return

`const [provided, snapshot] = useDraggable({ id, index, })`

that way you could easily rename, and it is similar to the way react core works

gnapse:
I wouldn't extrapolate that feature of `useState` to this case. I'm pretty sure `useState` does it like that to make it easier to rename the two things to your liking, since is common to use more than one state in several situations.

With something like this hypothetical `useDraggable`, that's not the case as much. Also, keep in mind there will be more stuff in that returned value, such as `isDragging`. What if you do not need everything? You'd need to remember the order of the stuff to know how to destructure it. With `useState` that's not an issue because is just two things and you pretty much always need them both.

YurkaninRyan:
@gnapse definitely, however the way I picture it, useDraggable would only return two things.  For example, check this from the docs.

![image](https://user-images.githubusercontent.com/8742831/47592232-75114300-d940-11e8-9dda-db919d105b72.png)

Here we get a `provided` object as well as a `snapshot` object.  Whether it returns an array or an object doesn't matter so much to me, just that we have access to both.

Unless you are suggesting all the properties get merged on one `draggable` return object, which is an interesting thought, but different from the current api i'm used to

alexreardon:
This seems like a reasonable API

```js
function Card({ itemId, index }) {
  const [provided, snapshot] = useDraggable({ draggableId: itemId, index });
  return (
    <div {...provided.draggableProps} {...provided.dragHandleProps } ref={provided.innerRef}>
      Drag me!
    </div>
  );
}
```

and 

```js
function List({ listId }) {
[...]

PR #1208, opened 2019-03-31
Hooks 🎣

alexreardon:
#871


.size-snapshot.json
```diff
@@ -1,25 +1,25 @@
...
+    "minified": 147989,
+    "gzipped": 41340
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 303784,
-    "minified": 113686,
-    "gzipped": 32910
+    "bundled": 327193,
+    "minified": 117669,
+    "gzipped": 33593
```
alexreardon: This was suprising

.size-snapshot.json
```diff
@@ -1,25 +1,25 @@
...
-    "minified": 125718,
-    "gzipped": 31356,
+    "bundled": 239473,
+    "minified": 123943,
+    "gzipped": 31513,
     "treeshaked": {
       "rollup": {
-        "code": 85891,
-        "import_statements": 832
+        "code": 30284,
```
alexreardon: This is quite a nice drop. @TrySound I am not sure what this property is referring too. Can you please explain this property?

.size-snapshot.json
alexreardon: I was expecting moving away from classes would be a drop, but it has gone the other way

src/view/droppable/droppable.jsx
```diff
@@ -1,191 +1,145 @@
...
+          innerRef={setPlaceholderRef}
+          animate={animate}
+          styleContext={styleContext}
+          onTransitionEnd={onPlaceholderTransitionEnd}
+        />
+      )}
+    </AnimateInOut>
+  );
+
+  // const placeholder: Node | null = instruction ? (
```
alexreardon: TODO: remove

src/view/placeholder/placeholder.jsx
```diff
@@ -48,129 +58,145 @@ const empty: Size = {
...
+    marginLeft: size.margin.left,
+
+    // ## Avoiding collapsing
+    // Avoiding the collapsing or growing of this element when pushed by flex child siblings.
+    // We have already taken a snapshot the current dimensions we do not want this element
+    // to recalculate its dimensions
+    // It is okay for these properties to be applied on elements that are not flex children
+    flexShrink: '0',
+    flexGrow: '0',
+    // Just a little performance optimisation: avoiding the browser needing
```
alexreardon: TODO: move to style marshal?

.size-snapshot.json
TrySound: This is the code in user bundle when you import nothing. Useful for ui libs like material ui. Does not mean a lot for this lib though.

src/view/drag-drop-context/use-startup-validation.js
```diff
@@ -0,0 +1,17 @@
...
+import checkDoctype from './check-doctype';
+
+export default function useStartupValidation() {
+  // Only need to run these when mounting
+  useEffect(() => {
+    if (process.env.NODE_ENV === 'production') {
+      return;
+    }
+
+    checkReactVersion(peerDependencies.react, React.version);
```
TrySound: I would wrap this check with condition. It looks like early returns does not eliminated well with webpack uglify pair.

alexreardon: Thanks

alexreardon: TODO: update

PR #1235, opened 2019-04-09
Fix broken link to Provider component

erik-beus:
The link to the react-redux `Provider` component is no longer valid.


alexreardon:
Thanks @erik-beus!!

PR #1236, opened 2019-04-09
11.0 🎣

### LINKED ISSUE #1193: 11.0 Breaking changes
Here is a list of planned breaking changes for `11.0`

#### Name change
shouldRespectForce**Touch** => shouldRespectForce**Press**

We will be calling preventDefault on the following: 
`webkitmouseforcechanged` (mouse) and `touchforcechange` (touch)

I think moving to `shouldRespectForcePress` is a clearer name given that it can be fired from mouse inputs too. `ForcePress` communicates the intent better

alexreardon:
Now that `react-redux@7` has gone stable - we should be good to go

Closes #1193


tommarien:
Just a heads up 11 is being tagged as latest and not as beta on npm ^^

alexreardon:
Looks like I missed this:

```bash
npm publish --tag beta
```

Next time!

Issue #1194, opened 2019-03-21
11.0 Planned breaking changes (not much!)


alexreardon:
Here is a list of planned breaking changes for `11.0`

### 1. React peer dependency

Moving from `^16.3.1` to `^16.8.0`. 

#### Reasons:

- Internally we will be moving things incrementally to hooks #871. 
- Need `16.8` to take advantage of `react-redux@7` https://github.com/reduxjs/react-redux/issues/1177
- Moving to `16.6` context API. Still using old school react context

#### 2.1 Name change

shouldRespectForce**Touch** => shouldRespectForce**Press**

We will be calling preventDefault on the following: 
`webkitmouseforcechanged` (mouse) and `touchforcechange` (touch)

I think moving to `shouldRespectForcePress` is a clearer name given that it can be fired from mouse inputs too. `ForcePress` communicates the intent better

#### 2.2 Flipping the default

See #1160. Opting for a better dragging experience out of the box

```diff
- shouldRespectForcePress: true,
+ shouldRespectForcePress: false,
```

alexreardon:
A huge thanks to @markerikson for his hard work on `react-redux` 👏

alexreardon:
Shipping #871

alexreardon:
```bash
react-beautiful-dnd@11.0.0-beta
```

You can have a play!

alexreardon:
`react-beautiful-dnd@11.0.0-beta.2` is out!

alexreardon:
`react-redux@7` has gone stable. I hope to ship `11.0` soon 😲

pwolaq:
Hi, I just want to report that #1138 is broken in `11.0.0-beta.2` - see here: https://codesandbox.io/s/ypv8xn70j

alexreardon:
Thanks for the heads up @pwolaq - i will take a look

alexreardon:
@pwolaq issue is now fixed in `11.0.0-beta.3`

alexreardon:
Released: https://github.com/atlassian/react-beautiful-dnd/releases/tag/v11.0.0

PR #1240, opened 2019-04-10
fixing focus loss in drag handle

alexreardon:
Closes issue raised by @pwolaq in #1194

### DIFF
--- a/src/view/use-drag-handle/use-focus-retainer.js
+++ b/src/view/use-drag-handle/use-focus-retainer.js
@@ -61,7 +61,7 @@ export default function useFocusRetainer(args: Args): Result {
         focusRetainer.retain(last.draggableId);
       }
     };
-  }, [getDraggableRef, lastArgsRef]);
+  }, [getDraggableRef]);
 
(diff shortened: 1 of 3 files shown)


src/view/use-drag-handle/use-focus-retainer.js
```diff
@@ -74,7 +74,7 @@ export default function useFocusRetainer(args: Args): Result {
     }
 
     // no change in ref
-    if (draggableRef === lastArgsRef.current.draggableId) {
+    if (draggableRef === lastDraggableRef.current) {
```
alexreardon: the fix

test/unit/view/drag-handle/focus-management.spec.js
```diff
@@ -99,6 +100,29 @@ describe('Portal usage (ref changing while mounted)', () => {
     }
   }
 
+  it('should not try to retain focus if the ref has not changed', () => {
```
alexreardon: test to ensure no regressions

PR #1264, opened 2019-04-17
moving to aliased import

alexreardon:


src/view/drag-drop-context/app.jsx
```diff
@@ -1,8 +1,9 @@
 // @flow
 import React, { useEffect, useRef, type Node } from 'react';
+import invariant from 'tiny-invariant';
 import { bindActionCreators } from 'redux';
 import { Provider } from 'react-redux';
-import { useMemoOne, useCallbackOne } from 'use-memo-one';
+import { useMemo, useCallback } from 'use-memo-one';
```
TrySound: Why? They may be confused with original react hooks.

alexreardon: It works nicer with `eslint-plugin-react-hooks` which matches on `useMemo` and `useCallback`

alexreardon: It is up to you though. You can still use `import { useMemoOne } from 'use-memo-one'`

TrySound: Ah, okay. Didn't think about lint rule.

PR #1294, opened 2019-05-06
Remove custom ExpectError from source code

TrySound:
ExpectError is not always defined in user config and flow fails there.

MutatableRefObject is just exact object with current field. I don't
think it will be exported.

```js
type MutatableRefObject<T> = {| current: T |};
```


alexreardon:
Love this!

PR #1322, opened 2019-05-25
Multi-drag example: fix incorrect selection count

stansv:
When dragging multiple items the selection count is also shown to display how many of these items in drag, but now this count includes even items selected in another column.

PS Another and IMO better way to fix this is to forcibly clear selection on 2nd list when drag is started/initial drop is finished, but I don't have a time right now to create such fix :(


alexreardon:
Nice!

alexreardon:
I will merge on monday

alexreardon:
Right now the count is done as it will allow you to drag items from multiple columns at the same time. If you want to disallow selection across columns you could do that in your selection logic.

Right now our example support dragging from multiple lists at a time, so the count should reflect that

PR #1340, opened 2019-06-04
Translate README.md to PT-BR

dudestein:
Translated the README.md file to Portuguese. Will look for additional resources needing translations as well.


alexreardon:
This looks great. But I think this should target a seperate repo and not master. See https://github.com/vtereshyn/react-beautiful-dnd-ru

PR #1347, opened 2019-06-07
using older theme

alexreardon:


alexreardon:
theme@5.x seems to be having issues

PR #1349, opened 2019-06-09
Resolve #1108 - error handling improvements

### LINKED ISSUE #1108: Improve how we handle errors
## Bug or feature request?

Vague, I'd consider it a bug.

### Expected behavior

React-beautiful-dnd doesn't touch errors that don't relate to it.

### Actual behavior

Instead of one error, I get 5 bulky errors that have nothing do with react-beautiful-dnd.
image

16.7.0.

10.0.3

Chrome.

thesuperhomie:
Keep track of errors with an internal store and add a type property to the errors we catch with the error boundary so we only keep track of internal errors.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 393181,
-    "minified": 147695,
-    "gzipped": 41532
+    "bundled": 394170,
+    "minified": 147884,
+    "gzipped": 41623
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 324803,
-    "minified": 116622,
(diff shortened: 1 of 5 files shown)


alexreardon:
Off to a good start!

Some errors are thrown outside of the react lifecycle - such as by *drag handle* window event handlers

thesuperhomie:
Ah didn't consider that, I'll try to get to that soon!

thesuperhomie:
@alexreardon I've thought about this a little bit and not sure what the best approach would be. What would you recommend?

alexreardon:
Thanks for thinking through this problem. We have done a lot of work to fix this up for `12.0`

alexreardon:
This is now improved in 12.0! Thanks for helping think through this

Issue #982, opened 2018-12-04
Move from Flow to TypeScript


alexreardon:
At Atlassian we consolidating on a single type system: `TypeScript`. 

In 2019 we plan on moving `react-beautiful-dnd` to `TypeScript`.

We still hope to have a good `flow` story through the `flowtyped` project. We can essentially just use our existing `flow` types for a `react-beautiful-dnd` library definition.

alexreardon:
For me personally this is not a negative reflection on `flow`. I really enjoy using it. However, there are advantages for us in consolidating on a type system internally. Also, my understanding is that `TypeScript` is a more popular project so a lot more people will get first class types when we move over ❤️

TrySound:
I would prefer you to keep flow types up-to-date in this repo. As you can see there's not so much people who will maintain them separately.

TrySound:
How about rewriting with reason? Type system is much nicer.

lilymatcha:
@alexreardon This is an exciting development! Since Typescript is more widely used, it will also help make it easier for people to contribute to the project. Within 2019, what is your expected timeline for this migration?

treshugart:
Just as we didn't want to maintain TS types when the source was in Flow, I don't think it's worth maintaining Flow types when moving to TS.

@lilymatcha Alex will be back mid-Jan. See #1010 for more details.

TrySound:
@treshugart Flow types will be maintained. There are existing users.

alexreardon:
@TrySound I am still thinking of the best way to maintain the flow types. Given they already exist I am open to having them in the repo. Once we start on the TypeScript conversion then we will have a better idea what the story could be

TrySound:
Types will be quite small. So it won't take a lot of efforts to maintain them. Keeping them separately will make them out of date soon.

mdestagnol:
@alexreardon good news for typescript. Do you guys also plan to move over Atlaskit?

nutboltu:
@alexreardon Great initiatives! As an initiative, any thoughts of moving `@types/react-beautiful-dnd` to this repos?

MaximSagan:
@alexreardon Has the migration to TS been abandoned?

eran-pinhas:
Hey guys, thought I would give it a shot

[...]

PR #1358, opened 2019-06-13
Add typescript types into repo

benhjames:
A nice first step toward TypeScript support (#982) would be to move the typings from DefinitelyTyped into this repo. This PR does just that! 🎉

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ flow-typed/
 site/
 coverage/
 babel.config.js
+typescript/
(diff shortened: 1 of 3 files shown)


alexreardon:
This seems like a reasonable short term solution

alexreardon:
Are the types from the existing @types repo?

benhjames:
Yep, it's an exact copy paste with the exception of 4 spaces -> 2 spaces indentation.

Ref: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-beautiful-dnd/index.d.ts

alexreardon:
For now I will hold off this one as the plan is to move the whole project over to typescript #982

PR #1363, opened 2019-06-17
Upgrade to core-js 3

petermikitsh:
Upgrade from `@babel/runtime-corejs2` to `@babel/runtime-corejs3`.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 393181,
-    "minified": 147695,
-    "gzipped": 41532
+    "bundled": 437376,
+    "minified": 160742,
+    "gzipped": 46554
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 324803,
-    "minified": 116622,
-    "gzipped": 33468
+    "bundled": 369034,
+    "minified": 129660,
+    "gzipped": 38532
   },
   "dist/react-beautiful-dnd.esm.js": {
-    "bundled": 239412,
-    "minified": 124303,
-    "gzipped": 31620,
+    "bundled": 244819,
+    "minified": 128055,
+    "gzipped": 32149,
     "treeshaked": {
       "rollup": {
-        "code": 30396,
-        "import_statements": 793
+        "code": 31532,
+        "import_statements": 1829
       },
       "webpack": {
-        "code": 34327
+        "code": 36069
       }
     }
   }
--- a/yarn.lock
+++ b/yarn.lock
@@ -1392,12 +1392,12 @@
     "@babel/helper-plugin-utils" "^7.0.0"
     "@babel/plugin-transform-typescript" "^7.3.2"
 
-"@babel/runtime-corejs2@^7.4.5":
+"@babel/runtime-corejs3@^7.4.5":
   version "7.4.5"
-  resolved "https://registry.yarnpkg.com/@babel/runtime-corejs2/-/runtime-corejs2-7.4.5.tgz#3d892f0560df21bafb384dd7727e33853e95d3c9"
-  integrity sha512-5yLuwzvIDecKwYMzJtiarky4Fb5643H3Ao5jwX0HrMR5oM5mn2iHH9wSZonxwNK0oAjAFUQAiOd4jT7/9Y2jMQ==
+  resolved "https://registry.yarnpkg.com/@babel/runtime-corejs3/-/runtime-corejs3-7.4.5.tgz#7df89702351c5fdc34b5555544fcf11b4ea7051c"
+  integrity sha512-yxHpUObdpx4zbuLJOHIZkw43E4NbWGsGH+85kRUj+VEcdtjlqI0vDcAja9mwOHqD7/4vCZp69q6eztz/sFDibQ==
   dependencies:
-    core-js "^2.6.5"
+    core-js-pure "^3.0.0"
     regenerator-runtime "^0.13.2"
 
 "@babel/runtime@7.3.4", "@babel/runtime@^7.0.0", "@babel/runtime@^7.1.2":
@@ -4504,6 +4504,11 @@ core-js-pure@3.1.3, core-js-pure@^3.0.1:
   resolved "https://registry.yarnpkg.com/core-js-pure/-/core-js-pure-3.1.3.tgz#4c90752d5b9471f641514f3728f51c1e0783d0b5"
   integrity sha512-k3JWTrcQBKqjkjI0bkfXS0lbpWPxYuHWfMMjC1VDmzU4Q58IwSbuXSo99YO/hUHlw/EB4AlfA2PVxOGkrIq6dA==
 
+core-js-pure@^3.0.0:
+  version "3.1.4"
+  resolved "https://registry.yarnpkg.com/core-js-pure/-/core-js-pure-3.1.4.tgz#5fa17dc77002a169a3566cc48dc774d2e13e3769"
+  integrity sha512-uJ4Z7iPNwiu1foygbcZYJsJs1jiXrTTCvxfLDXNhI/I+NHbSIEyr548y4fcsCEyWY0XgfAG/qqaunJ1SThHenA==
+
 core-js@3.0.0:
   version "3.0.0"
   resolved "https://registry.yarnpkg.com/core-js/-/core-js-3.0.0.tgz#a8dbfa978d29bfc263bfb66c556d0ca924c28957"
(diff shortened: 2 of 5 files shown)


petermikitsh:
One thing not addressed in this PR, but might be valuable long-term: I've noticed that I'm getting duplicated copies of core-js things. Ideally, the imports would be de-duplicated.

In my project, here's some of my babel config:

```js
      [
        "@babel/preset-env",
        {
          "modules": false,
          "corejs": 2,
          "useBuiltIns": "usage"
        }
      ],
```

In my transpiled source code, this leads to imports like:

```js
import "core-js/modules/es6.object.assign";
```

Conversely, react-beautiful-dnd will import:

```js
import _Object$assign from '@babel/runtime-corejs2/core-js/object/assign';
```

Which points to ([source](https://unpkg.com/@babel/runtime-corejs2@7.4.5/core-js/object/assign)):

```js
module.exports = require("core-js/library/fn/object/assign");
```

`core-js/modules/es6.object.assign` !== `core-js/library/fn/object/assign`.

Bytes are getting duplicated, and I'm unsure how to resolve it.

---

react-beautiful-dnd's import:

Screen Shot 2019-06-17 at 4 15 41 PM

my import:

Screen Shot 2019-06-17 at 4 15 26 PM

alexreardon:
How does this look to you @TrySound ?

.size-snapshot.json
```diff
@@ -1,25 +1,25 @@
...
+    "minified": 160742,
+    "gzipped": 46554
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 324803,
-    "minified": 116622,
-    "gzipped": 33468
+    "bundled": 369034,
+    "minified": 129660,
+    "gzipped": 38532
```
alexreardon: ouch! this is a 9% change

alexreardon:
Can you look a bit more into core-js duplicates too? Let's do the right thing here.

The change in this PR looks fine. However, it is a 10% hit on our `min` bundle

yarn.lock
```diff
@@ -4504,6 +4504,11 @@ core-js-pure@3.1.3, core-js-pure@^3.0.1:
   resolved "https://registry.yarnpkg.com/core-js-pure/-/core-js-pure-3.1.3.tgz#4c90752d5b9471f641514f3728f51c1e0783d0b5"
   integrity sha512-k3JWTrcQBKqjkjI0bkfXS0lbpWPxYuHWfMMjC1VDmzU4Q58IwSbuXSo99YO/hUHlw/EB4AlfA2PVxOGkrIq6dA==
 
+core-js-pure@^3.0.0:
```
alexreardon: core-js-pure?

.size-snapshot.json
alexreardon: Relevant? 

- https://github.com/zloirock/core-js/blob/master/docs/2019-03-19-core-js-3-babel-and-a-look-into-the-future.md
- https://github.com/zloirock/core-js/blob/master/README.md#configurable-level-of-aggressiveness

alexreardon: I feel like we are not doing something right to get this size bump

yarn.lock
petermikitsh: That's correct. `@babel/runtime-corejs3` uses `core-js-pure`:

https://github.com/babel/babel/blob/33ab4f166117e2380de3955a0842985f578b01b8/packages/babel-runtime-corejs3/package.json#L11-L14

The old version (`@babel/runtime-corejs2`) used `core-js`:

https://github.com/babel/babel/blob/33ab4f166117e2380de3955a0842985f578b01b8/packages/babel-runtime-corejs2/package.json#L11-L14

petermikitsh:
I've done a bit more research here. I'm working under the assumption that most people are also using `@babel/preset-env` in their own projects, in conjunction with this library. As a recap, `preset-env` transpiles ES6 based off a [browserslist](https://github.com/browserslist/browserslist). For polyfills, it uses `core-js` configured via the [useBuiltIns](https://babeljs.io/docs/en/babel-preset-env#usebuiltins) option. You have the choice of specifying core-js [v2 or v3](https://babeljs.io/blog/2019/03/19/7.4.0#migration-from-core-js-2) -- by default, it's v2.

With this change set, all of your `react-beautiful-dnd` polyfills come from `core-js-pure`. Polyfills from preset-env are from `core-js`.

To remove duplicates, you can alias `core-js` to `core-js-pure` in Webpack like so:

```js
{
  resolve: {
    alias: {
      'core-js': 'core-js-pure'
    }
  }
}
```

---

I compared my bundle size using various setups outlined below.

### Using v11.0.4, @babel/preset-env configured to core-js@2 (without aliasing)

264 kB

### Using v11.0.4, @babel/preset-env configured to core-js@3 (without aliasing)

271 kB

### Using this PR Build, @babel/preset-env configured to core-js@3 (without aliasing)

287 kB

### Using this PR Build, @babel/preset-env configured to core-js@3 (with aliasing)

260 kB

.size-snapshot.json
petermikitsh: https://babeljs.io/blog/2019/03/19/7.4.0#migration-from-core-js-2

I've read through the migration docs for the 7.4.0 release. It basically says to do this:

package.json:

```diff
- @babel/runtime-corejs2
+ @babel/runtime-corejs3
```

babel config:

```diff
plugins: [
  ["@babel/transform-runtime", {
-    corejs: 2,
+    corejs: 3,
  }]
]
```

If anything looks missed, let me know -- but I'm thinking size bump is correct.

alexreardon: I wonder if we need to change something. Where is a near 10% increase coming from?

TrySound:
In a lot of libraries core-js is not used. Polyfilling is left for user. This let modern projects without IE support to reduce their bundles significantly.

alexreardon:
How do you think we should move forward @TrySound ?

TrySound:
Do you run automated tests in ie?

petermikitsh:
My 2 cents: Lots of popular libraries require users to specify their own polyfills, e.g., [React](https://reactjs.org/docs/javascript-environment-requirements.html). If you keep things un-opinionated, as @TrySound suggested, everyone can custom tailor their bundle size to what they need. People not targeting IE 11 probably won't need to worry about polyfills, for example. People who are can also choose whether they want to polyfill using core-js 2 or 3 or some other solution, which should address getting duplicate copies of polyfills. It leaves a lot of choice, but would be a breaking change, though.

alexreardon:
I didn't think we were polyfilling anything right now

TrySound:
This is from stable release

image

alexreardon:
My understanding is that none of those polyfill anything. My understanding is that they are ponyfills

petermikitsh:
That's correct, they are ponyfills. I think @TrySound's rationale for excluding polyfills still applies to ponyfills. Some users of the library could not be targeting older browsers (e.g., IE 11) so they would not need the ponyfill.

alexreardon:
Updating to corejs 3 makes sense. I am just concerned that one of our bundles went up 10%. Are we ponyfilling things that we were not before? Have the ponyfills gotten bigger?

petermikitsh:
Here's the diff for corejs v2 -> v3 esm build. Nothing particularly stands out to me. There's a lot more imports at the top-- that would account for some of the bundle size increase.

image

alexreardon:
That is super helpful thank you

alexreardon:
It feels like it is doing too much right now.

Why is it swapping out `.map` and `.reduce`, `.forEach` and `.filter`, `.slice()`!?. We are targetting es5; i don't think we should be modifying these. 

It is also being too aggressive with `.trim()` and all the functions in `native-with-fallback`. In there we provide our own light ponyfills conditionally.

petermikitsh:
A bit more info about `@babel/plugin-transform-runtime`... it ponyfills less with core-js [v2](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-runtime/src/runtime-corejs2-definitions.js) than [v3](https://github.com/babel/babel/blob/master/packages/babel-plugin-transform-runtime/src/runtime-corejs3-definitions.js#L187-L229). Also, `plugin-transform-runtime` doesn't consider target environments (as mentioned in https://github.com/babel/babel/issues/10008#issue-446717469). 

That's why we're observing the ponyfilling of things like `.forEach`, which definitely does not need a ponyfill, based off the environments listed in `.browserslistrc`. (All evergreen browsers plus IE 11 support Array.prototype.forEach natively).

With `plugin-transform-runtime`, Ponyfills are injected for every used ECMAScript feature supported by core-js. If you look closely at https://unpkg.com/react-beautiful-dnd@11.0.4/dist/react-beautiful-dnd.esm.js, you'll see ponyfills that are not needed. E.g., `Date.now` and `Object.keys` are currently ponyfilled in `v11.0.4`.

It might be a good idea to revisit this once https://github.com/babel/babel/issues/10008 shakes out a bit more.

alexreardon:
I am keen to get this in. It seems like https://github.com/babel/babel/issues/10008 is a blocker right now

PR #1371, opened 2019-06-21
Fix inconsistent language link styling in README.md

timhaywood:
update language links to use consistent text style (bold)


alexreardon:
Awesome!

PR #1383, opened 2019-06-27
Seancurtis/accessibility improvements

### LINKED ISSUE #1288: Improved JAWS and NVDA support
### Expected behavior
With Screen reader, movement of draggable item with arrow keys does not work.

### Actual behavior
Screen reader (Tested with JAWS and NVDA) arrow keys defaults to read next character. Because of this draggable items content get read instead of moving item to next position.

### Steps to reproduce
Open screen reader
navigate to example
press space bar for draggable item and press arrow keys
### Suggested solution?
in a drag mode (pressing of space bar on draggable item) capture arrow keys and prevent default behaviour

16.8.4

11.0.0

chrome 73.0.3683.103
IE11

### Demo
this reproducible on the stories we have on the examples link
https://react-beautiful-dnd.netlify.com/iframe.html?id=single-vertical-list--basic

This is an awesome library with keyboard support. Thanks for it. 
Extending it to be functional with screen readers will widen its impact exceptionally.

seancurtis:
Fixes #1288

### DIFF
--- a/src/view/drag-drop-context/drag-drop-context.jsx
+++ b/src/view/drag-drop-context/drag-drop-context.jsx
@@ -12,6 +12,7 @@ type Props = {|
 
   sensors?: Sensor[],
   enableDefaultSensors?: ?boolean,
+  dragHandleDescription?: string,
 |};
 
 let instanceCount: number = 0;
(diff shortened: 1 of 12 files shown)


src/view/drag-drop-context/drag-drop-context.jsx
```diff
@@ -21,15 +22,24 @@ export function resetServerContext() {
   instanceCount = 0;
 }
 
+const DEFAULT_DRAG_HANDLE_DESCRIPTION = `Draggable item. Ensure you're screen reader is not in browse mode and then press spacebar to lift.`;
```
alexreardon: wrong "you're"

alexreardon: We will also need to update the screen reader guide lift announcement section

alexreardon: I will fix

PR #1386, opened 2019-06-27
add a11y auditing

stacylondon:


.gitignore
```diff
@@ -17,8 +17,8 @@ site/
 # coverage reports
 coverage/
 
-# test reports
-test-reports/
+# a11y reports
+a11y-audit.json
```
alexreardon: can we put this with our other reports?

.circleci/config.yml
```diff
@@ -133,6 +133,25 @@ jobs:
...
+            - v7-dependencies-{{ checksum "yarn.lock" }}
+
+      # PR's from forks cannot use the dependency cache for performance reasons
+      - run:
+          name: 'Forked PR dependency install'
+          command: yarn
+
+      - run:
+          name: Accessibility Audit
+          command: node browser-test-harness.js yarn test:accessibility
```
alexreardon: can we get the json output added as an artifact?

stacylondon: done!

.gitignore
stacylondon: done!

package.json
```diff
@@ -33,6 +33,8 @@
     "prettier_target": "*.{js,jsx,md,json} src/**/*.{js,jsx,md,json} test/**/*.{js,jsx,md,json} docs/**/*.{js,jsx,md,json} stories/**/*.{js,jsx,md,json} cypress/**/*.{js,jsx,md,json}"
   },
   "scripts": {
+    "test:accessibility": "lighthouse http://localhost:9002/iframe.html?id=single-vertical-list--basic --chrome-flags='--headless' --output=json --output-path=./test-reports/lighthouse/a11y-audit.json && yarn test:accessibility:report",
+    "test:accessibility:report": "node a11y-audit-parse.js",
```
alexreardon: does this need to be a separate step?

stacylondon: At first I was thinking I might have `test:accessibility` and `test:accessibility:ci` and then they could share that common step but now there is only a need for `test:accessibility` so I'll consolidate and just have the one NPM script for this.

stacylondon: Done; just a single step now.

Issue #1401, opened 2019-07-11
forcepress on iOS


alexreardon:
Right now, if you call `event.preventDefault()` on a `touchforceevent` it will opt out of a lot of standard browser behaviours including clicking a scrolling.

I suspect this is a recent change in webkit, and I have [raised a bug on their project](bugs.webkit.org/show_bug.cgi?id=199658). Unfortunately, we cannot just wait for them, we need to find a path forward, even if it is only temporary.

Right now it is a delicate dance to enable drag and drop interactions in force press compatible environments.

### Background

Force press interactions on iOS allow for additional interactions on elements. The most notable is links prevents and navigation by applying different levels of pressure to an element

### `shouldRespectForcePress={false}` (current default)

This allows users to disable force press interactions and allow dragging with any pressure applied.

Call `event.preventDefault()` for all `touchforceevent` events after `touchstart`

💥This is now breaking scrolling #1396 and clicking #1367 on iOS.

### `shouldRespectForcePress={true}` (`10.x` default)

With this option we try to respect standard force press interactions as much as possible. We moved away from this as the default as people trying to drag might apply too much pressure and have their pending drag cancelled and not know why.

- Cancel a pending if we detect a force press (`touchforceevent` with enough pressure)
- If dragging and the user has not moved at all: we cancel the drag and allow the force press interaction
- If dragging and the has moved the draggable: then we call `event.preventDefault()` on `touchforceevent`s to prevent force press actions mid-drag.

### Workaround

I have tried a number of potential workaround strategies. The only one that seems to be viable is this:

- only call `event.preventDefault()` on `touchforceevent` after a force press is recognised (once enough pressure is registered). This is rather than doing it on every event
- reduce the time to lift from `150ms` to `120ms`. This seems to be safely inside of the time that webkit gives for registering a force press interaction. This is not ideal as I have noticed it can interfere with slightly slower click interactions

I think this is the best path forward for now. Hopefully, webkit will action the bug I raised and we can change things back to how they were

## TODO:

- [x] Verify workaround works and has no significant gotchas
- [x] Investigate `webkitmouseforcechanged` and if it needs the same workaround
- [x] Write tests
- [x] Backport fix from `12.x` `alpha` to `11.x`

alexreardon:
I have shipped the workaround in an alpha for validation: 

`12.0.0-alpha.7`

alexreardon:
[...]

Issue #1367, opened 2019-06-19
Cannot get onClick event to fire on a draggable element in latest ios safari


jimmybillings:
### Expected behavior
I should be able to add an onClick handle to a draggable element and see it fire when I click on the draggable element without dragging it. 

### Actual behavior
Almost never fires the click event. If I keep persisting it will. 

### Steps to reproduce
Visit the example for a simple list on an ios device using safari - https://react-beautiful-dnd.netlify.com/iframe.html?id=single-vertical-list--basic

On a desktop if you click one of the items it will link you to another url, but on a ios mobile using safari if you touch one of the items it doesn't go anywhere. I haven't confirmed this is only with IOS and safari and not all mobile. I'm also wondering if there is any correlation with this and the following issue which I'm also experiencing: https://github.com/atlassian/react-beautiful-dnd/issues/1314

alexreardon:
Is this an issue?

jimmybillings:
Potentially, I'm not sure how it's intended as I didn't find a ton of documentation on using onClick with dnd on a mobile device. 

I moved to onTouchEnd with some prevent default logic to make it work on mobile that works but using onClick never seemed to work, not just safari but all browsers. I've never run into this before doing responsive sites with react, onClick has always seemed to work fine but this solution is fine for me with onTouchEnd if that's how you designed it.

alexreardon:
My expectation is that clicking should be unaffected by `rbd`. Let me know how you go

jimmybillings:
Ok. After more testing onClick is definitely not working on ios with all browsers, at least in my case. Seems to work fine on android. I will put together a code sample on stackblitz to reproduce shortly. Thanks.

alexreardon:
Thanks!

jimmybillings:
Ok I have a code example. I have copied the simple list example and added an onClick handle to the element inside of Draggable that creates a browser alert. Works fine in the desktop browser but not in ios. Of course you will need to look at this from an iphone/ipad to view the potential bug. 

App url here: https://react-ub51sd.stackblitz.io
Editor Url here: https://stackblitz.com/edit/react-ub51sd

alexreardon:
Before I dig into it more, can you please try using `react-beautiful-dnd@12.0.0-alpha.4` and see if that fixes the issue?

jimmybillings:
I gave it a go but no luck unfortunately, same deal.

alexreardon:
[...]

PR #1407, opened 2019-07-15
back porting force press fix from virtual branch

alexreardon:


test/unit/view/drag-handle/mouse-sensor.spec.js
```diff
@@ -1233,170 +1233,3 @@ describe('subsequent drags', () => {
     ).toBe(true);
   });
 });
-
-describe('webkit force press', () => {
```
alexreardon: removing tests as `virtual` has tests for this new behaviour

alexreardon:
Closes #1401 #1367

PR #1481, opened 2019-09-09
chore(testcafe): Removing testcafe dead codes from the repo

### LINKED ISSUE #1480: Removing testcafe packages
After reviewing this PR https://github.com/atlassian/react-beautiful-dnd/pull/1122, it seems like this library has already migrated from **testcafe** to **cypress**.

I found this package (**testcafe-reporter-xunit**) https://github.com/atlassian/react-beautiful-dnd/blob/master/package.json#L132 in the package.json which should be removed from this library.

There are also a few places where I found dead codes for testcafe. We can remove those as well.

nutboltu:
Fixes #1480 

This PR includes

- Remove package `testcafe-reporter-xunit`
- Remove flow-typed for `testcafe`


alexreardon:
Nice!

PR #1460, opened 2019-08-21
add isCombineOnly alternative to isCombineEnabled

mattkrick:
Fixes #1390
simply pass `isCombineOnly` instead of `isCombineEnabled` to turn off reordering.

Issue #1390, opened 2019-07-04
Allow combining without reordering


alexreardon:

mattkrick:
hey @alexreardon happy to attempt a PR for this one. i noticed there was already an attempt that went stale (#1129).

Before i get started any guidance on general architecture that you'd like to see?

mattkrick:
closed in favor of #1505

DanielPower:
Re-posting my suggestion from #1506 here.

Currently we have `isCombinedEnabled`, which allows you to set whether or not items can be dragged on top of each other to be combined. And we have `isDropDisabled`, which prevents items from being dropped onto the `Droppable` entirely.

I think we can solve this issue (combining without reordering) without adding more props or complexity. My proposal is to replace `isDropDisabled` with `isReorderEnabled`. Then combining and reordering can be controlled individually. Disabling both will have the same behaviour as `isDropDisabled={true}` does now.

The result would have these four states, depending on the values of `isCombineEnabled` and `isReorderEnabled`:

    `isCombineEnabled={true} isReorderEnabled={true}`:
        The list can be reordered, and items can be dropped onto one another. Behaves exactly the same as the current behaviour with `isCombineEnabled={true} isDropDisabled={false}`.
    `isCombineEnabled={false} isReorderEnabled={true}`:
        The same as the default behaviour.
    `isCombineEnabled={true} isReorderEnabled={false}`:
        Items can be picked up, and dropped onto one another, but will not reorder. Items will not animate to move out of the way. This behaviour is not possible currently, as far as I've been able to find.
    `isCombineEnabled={false} isReorderEnabled={false}`:
        Works the same as `isDropDisabled={true}`

Adding an item to the list from another list, would also be considered "reordering", and would be disallowed in `isReorderEnabled={false}`.

philipaarseth:
Is there any chance we could see an update on this? 
It seems like people have been wanting a similar feature like this since early 2018.

mattkrick:
@philipaarseth FWIW, it's really, truly easier to build your own package to support this. this package has some beautiful patterns in place to learn from (namely snubbing HTML DnD for mousemove events, support for mobile), but it will only get you 80%. the rest (namely perfect animations & ungrouping) will drive you mad. 

[...]

PR #1505, opened 2019-09-19
Allow combine without reordering

GDIBass:
Implements: Issue #1390

This mostly piggybacks off of what @mattkrick started here: https://github.com/atlassian/react-beautiful-dnd/pull/1460 , except I believe I've fixed the testing issues.

I'm not well versed in Jest, so I haven't built out any tests yet.  If anyone can help with that let me know, otherwise I'm going to give it a shot tomorrow.

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,18 +1,18 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 394771,
-    "minified": 148402,
-    "gzipped": 41581
+    "bundled": 394739,
+    "minified": 148398,
+    "gzipped": 41586
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 326088,
-    "minified": 117128,
-    "gzipped": 33569
+    "bundled": 326056,
+    "minified": 117124,
+    "gzipped": 33573
   },
   "dist/react-beautiful-dnd.esm.js": {
-    "bundled": 241310,
-    "minified": 125511,
-    "gzipped": 31738,
+    "bundled": 241279,
+    "minified": 125507,
+    "gzipped": 31745,
     "treeshaked": {
       "rollup": {
         "code": 30265,
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 392081,
-    "minified": 147079,
-    "gzipped": 41340
+    "bundled": 394771,
+    "minified": 148402,
+    "gzipped": 41581
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 323864,
-    "minified": 116189,
-    "gzipped": 33372
+    "bundled": 326088,
+    "minified": 117128,
+    "gzipped": 33569
   },
   "dist/react-beautiful-dnd.esm.js": {
-    "bundled": 238384,
-    "minified": 123773,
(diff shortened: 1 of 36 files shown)


DanielPower:
Sorry if I'm missing something obvious, but could you tell me how test your branch? I've pulled it down with `npm i GDIBass/react-beautiful-dnd#is-combine-only`, but the resulting `node_modules/react-beautiful-dnd` does not contain a dist directory. Just src and 

I'm not familiar with the build process. Is there a command I need to run to generate the dist directory?

GDIBass:
@DanielPower my own testing process was as follows:

1. Install the package on a project of mine
2. Download this repo
3. Run yarn install
4. run yarn build
5. Copy the dist/src files to node_modules on my own project
6. Build my project
7. Test ahoy.

I'm pretty sure there's a better way to do it using the tarball method here, but the way I did it was decently easy.   https://github.com/atlassian/react-beautiful-dnd/pull/1460

I haven't had a chance to look into building out more tests to cover these, but I will shortly.  I don't think tonight, but likely over the weekend.

DanielPower:
Thanks for the instructions, that worked perfectly. I didn't know about yarn until now. Also a minor suggestion, you could simply symlink your node_modules/react-beautiful-dnd/dist directory to the dist directory in your build directory, to save you from having to re-copy every time.

As for the PR, I just tested it out, and it's really close! But there's a bit of an issue with it (unless I'm doing something wrong). Attached is a gif of me attempting to use `isCombineOnly`.

As you can see, although reordering is disabled, the space that was taken up by the item being dragged does get filled in, and it seems the position of items do not get updated in the combine logic. 

video of issue

GDIBass:
Good find.  I'll take a look at that over the weekend.

Definitely would like @alexreardon 's input there.  I personally like the "isCombineOnly" prop, because it's intuitive to use.

ArielFrischer:
Would love to see this implemented I've tried multiple workarounds in other issues to no avail because placeholder seems to keep moving around behind the scenes. Any updates on whether this will get merged?

dannywitters:
any updates on when this might get merged?

MichielMortier:
Could this be merged?

spideyboy18:
When will it be merged?

GDIBass:
I ended up abandoning this for a different solution.  If anyone wants to pick it up and run with it, feel free.  As is it definitely doesn't pass all required tests.

matheusAle:
any update about this?

ariel-frischer:
> I ended up abandoning this for a different solution. If anyone wants to pick it up and run with it, feel free. As is it definitely doesn't pass all required tests.

Can you explain the new solution you have used?

GDIBass:
For my use case I ended up setting up individual drop containers for each of the slots, so instead of doing

drop container with items: [a, b, c, d, e]

I ended up with:
[a] [b] [c] [d] [e]

You'd drag from e to a, then I'd update the a drop container with e

[ae] [b] [c] [d] []

May not work for everybody but for my use case it got the job done pretty well

ariel-frischer:
> For my use case I ended up setting up individual drop containers for each of the slots, so instead of doing
> 
> drop container with items: [a, b, c, d, e]
> 
> I ended up with:
> [a] [b] [c] [d] [e]
> 
> You'd drag from e to a, then I'd update the a drop container with e
> 
> [ae] [b] [c] [d] []
> 
> May not work for everybody but for my use case it got the job done pretty well

Sounds good thanks!

PR #1578, opened 2019-10-31
Improving the accessibility for lift instructions

seancurtis:


src/view/use-lift-instruction/use-lift-instruction.js
```diff
@@ -24,8 +23,8 @@ export default function useLiftInstruction(
       // add the description text
       el.textContent = liftInstruction;
 
-      // hide the element visually
-      Object.assign(el.style, visuallyHidden);
```
alexreardon: Ah okay. So there are some things we want visually hidden, but available to the screen readers. And there are other things we want visually hidden + not available to screen readers

test/unit/integration/accessibility/axe-audit.spec.js
```diff
@@ -0,0 +1,20 @@
+// @flow
```
alexreardon: 👏

src/view/use-lift-instruction/use-lift-instruction.js
```diff
@@ -24,8 +23,8 @@ export default function useLiftInstruction(
...
+      // prevent AT from idly reading it. Fixes #1568
```
alexreardon:
```suggestion
// Visually hide the element and prevent screen readers from reading this element in the document flow
// This element is used as a `aria-labelledby` reference for *other elements* and will be read out for those
```

src/view/use-lift-instruction/use-lift-instruction.js
```diff
@@ -24,8 +23,8 @@ export default function useLiftInstruction(
...
+      Object.assign(el.style, { display: 'none' });
```
alexreardon: Can you confirm that the lift instructions still work as expected?

src/view/use-lift-instruction/use-lift-instruction.js
seancurtis: Yeah if we reference it only via something like `aria-describedby` then it doesn't need to be visible. The `visuallyHidden` is mainly for things like icon buttons (using `aria-label` is better), and I believe it's needed for live regions too.

src/view/use-lift-instruction/use-lift-instruction.js
seancurtis: Works fine in VO (90% sure it also works in NVDA/JAWS too, but didn't test)

alexreardon:
It looks like the flow violation is minor: it doesn't understand the jest extension. We can add a `//$FlowFixMe` for now. Then we can get this in :)

test/unit/integration/accessibility/axe-audit.spec.js
```diff
@@ -0,0 +1,20 @@
...
+
+expect.extend(toHaveNoViolations);
+
+// Need to investigate this further as it doesn't seem to pick up the same errors
+// which are found by react-axe.
+it('should not fail an aXe audit', async () => {
+  render(<App />);
+
+  const results = await axe(document.body);
+
```
alexreardon:
```suggestion
// $FlowFixMe - flow doesn't know about hte custom validator
```

test/unit/integration/accessibility/axe-audit.spec.js
```diff
@@ -0,0 +1,21 @@
...
+// $FlowFixMe - flow doesn't know about hte custom validator
```
alexreardon:
```suggestion
  // $FlowFixMe - flow doesn't know about hte custom validator
```

src/view/use-lift-instruction/use-lift-instruction.js
```diff
@@ -24,8 +23,9 @@ export default function useLiftInstruction(
       // add the description text
       el.textContent = liftInstruction;
 
-      // hide the element visually
-      Object.assign(el.style, visuallyHidden);
+// Visually hide the element and prevent screen readers from reading this element in the document flow
+// This element is used as a `aria-labelledby` reference for *other elements* and will be read out for those
```
alexreardon:
```suggestion
  // Visually hide the element and prevent screen readers from reading this element in the document flow Object.assign(el.style, visuallyHidden);
  // This element is used as a `aria-labelledby` reference for *other elements* and will be read out for those
```

Issue #930, opened 2018-11-15
Improve reordering with hugely mixed size Draggables


rd-dev-ukraine:
## Bug or feature request?

Bug. 

In case of mixing large-size draggables and small-size draggables we got a crazy behaviour.

### Expected behavior

I can properly rearrange large draggables and this should look nice (like on small draggables)

### Actual behavior

In my real application - I can't rearrange large draggables. Medium-sized draggables can be rearranged only if user will cross another item with center of current draggable. The problem is very huge, because only part of draggable does have dragHandleProps.
In my sample - it just looks crazy and it's non-trivially to rearrange large draggable, but possible under specific user action.

### Steps to reproduce

1. Run demo link;
2. Try to move "Item 2" (large block) between "Item 0" and "Item 1" in one action (it's impossible)
3. Try to move "Item 2" before "Item 0" in one action (sometime it works, sometime it doesn't)

### What version of `React` are you using?

16.4.1

### What version of `react-beautiful-dnd` are you running?

10.0.0

### What browser are you using?

Chrome (latest stable)

### Demo

https://codesandbox.io/s/qv629nm036

### Thoughts

It looks like the problem happens because react-beautiful-dnd uses center of draggable as control point. 
This is especially incorrect in case of drag handles. And even without drag handles it's better to use  mouse position inside of draggable as control point

alexreardon:
We use the center position as the control point as it is the items center of mass. 

But the example you are showing does have some lame interactions. So it is worth looking into a bit more

gabrielenosso:
I have a similar problem:
I have the same object that can be rendered in 3 different sizes.
In the destination, they are all shown as small.
When I drag it from the source, independently from the size, I show it as small.

So the problem is that it looks weird if I drag the Big object and then I show it as Small during the drag operation. (the position relative to the mouse cursors is broken)

gihrig:
Similar issue, more discussion at #978

gihrig:
@alexreardon How difficult would it be to create a config item to place the active point of a draggable at its "center of gravity" vs at the drag handle?

Jannikdam:
**Expected behavior**
An element "A" should move up or down across the element that is being dragged,"B", when the (upper or lower) border of B crosses the center of A.
This would be opposite of how draggables moved, before version 10:
[...]

PR #1603, opened 2019-11-13
Handling mixed size draggables when reordering

caspersmith:
#930


caspersmith:
@alexreardon turns out the original proposed approach didn't play well with virtualised lists. It relied on comparing array indicies to determine whether the item being dragged was before or after other items in the list. In a virtualised list we only view and process a small window of the items in the array and so this didn't work.

I've pushed a few more changes now to update the approach. It's a little more similar to what was already in place, but the logic has been tweaked slightly so that it compares the start/end points of the dragged item against the central position of other items in the list. Also no longer relies on the movement direction.

Changes have also been made to the checks around whether items are combined. It now compares in the same way as the reorder function, looking at whether the start/end of the dragged item has crossed into the central area of another item in the list.

This approach works reasonably well now in most of the cases I've tested, including with combine enabled and virtual lists.

alexreardon:
I updated the branch to pull in #1618 for comparision

alexreardon:
Something positive I have noticed about your approach is that it does not consider user direction. I think that is making the issue with big items worse right now. I am going to have a go at moving the existing logic over to not use the user direction. That way we can make a better comparision between the two approaches

PR #1606, opened 2019-11-14
correct aria-label for lift button

coorasse:


alexreardon:
This might seem like a small change, but I really appreciate it. Thanks for your time

PR #1611, opened 2019-11-14
Fix small typo on docs

igorrocha:
Update from _then_ to _than_ on "We cannot give more information **then** this because things might change"


alexreardon:
Cheers!

PR #1614, opened 2019-11-15
Change jest-axe dependency to dev

dominykas:
`jest-axe` is only used during testing, therefore it belongs in `devDependencies`.

This is normally not a big deal, aside from increased install time, but still.


alexreardon:
Thanks for the pickup!

PR #1633, opened 2019-11-26
remove event.preventDefault in mousedown event

robberfree:
1. when you click a normal element,then make it **contentEditable** and **focus** it.
Default behavior in most browsers it should be put cursor in right position and can edit the element's content.
2. If set **event.preventDefault()** in **mousedown** event.The default behavior will fly out.
3. I create a demo here:
https://codesandbox.io/s/click-then-contenteditable-react-xjkbx?fontsize=14&hidenavigation=1&theme=dark


robberfree:
I see two unit tests to make sure that **prevent the element gaining focus on the initial mousedown event**.
So is it possible to provide a on/off option to do task like this?
Thx.

PR #1636, opened 2019-11-27
Update useIsomorphicLayoutEffect check

### LINKED ISSUE #1623: useLayoutEffect warning during SSR (improve useIsomorphicLayoutEffect check)
### Expected behavior
There is a warning during SSR because window is defined, I would expect there to be no warning.

### Suggested solution
Update use-isomorphic-layout-effect to check for window, window.document and window.document.createElement instead of just window. 
That would match what both [React](https://github.com/facebook/react/blob/master/packages/shared/ExecutionEnvironment.js) and [Redux](https://github.com/reduxjs/react-redux/blob/master/src/utils/useIsomorphicLayoutEffect.js) are doing

### What version of React are you using?
16.8.6

### What version of react-beautiful-dnd are you running?
12.0.0

Chrome

LinusCenterstrom:
fixes #1623

Updated the the useIsomorphicLayoutEffect layout effect to check for window.document.createElement instead of just window.
That matches what both [React](https://github.com/facebook/react/blob/master/packages/shared/ExecutionEnvironment.js) and [Redux](https://github.com/reduxjs/react-redux/blob/master/src/utils/useIsomorphicLayoutEffect.js) are doing

### DIFF
--- a/src/view/use-isomorphic-layout-effect.js
+++ b/src/view/use-isomorphic-layout-effect.js
@@ -9,6 +9,10 @@ import { useLayoutEffect, useEffect } from 'react';
 // `connect` to perform sync updates to a ref to save the latest props after
 // a render is actually committed to the DOM.
 const useIsomorphicLayoutEffect =
-  typeof window !== 'undefined' ? useLayoutEffect : useEffect;
+  typeof window !== 'undefined' &&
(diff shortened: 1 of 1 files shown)


alexreardon:
Thanks for this! I hope to do a release late next week

PR #1637, opened 2019-11-27
Remove duplicate code in dimension locking example

CraigEge:
The same boolean check is performed twice, once after the other.

This cleans up the example for readability and to fix redundancy.

### DIFF
--- a/stories/src/table/with-dimension-locking.jsx
+++ b/stories/src/table/with-dimension-locking.jsx
@@ -191,10 +191,7 @@ export default class TableApp extends Component<AppProps, AppState> {
     });
 
     // dropped outside the list
-    if (
-      !result.destination ||
-      result.destination.index === result.source.index
-    ) {
+    if (!result.destination) {
       return;
     }
 


stories/src/table/with-dimension-locking.jsx
```diff
@@ -191,10 +191,7 @@ export default class TableApp extends Component<AppProps, AppState> {
     });
 
     // dropped outside the list
-    if (
-      !result.destination ||
-      result.destination.index === result.source.index
```
alexreardon: It is possible to move an item back into the same spot. The check ` result.destination.index === result.source.index` will exit early if that is the case

CraigEge: If you check the lines just below it, the same check occurs, making it redundant. https://github.com/atlassian/react-beautiful-dnd/blob/8e4c3845e8d396e8baf5d95c6077d49f09b9913b/stories/src/table/with-dimension-locking.jsx#L199

alexreardon: Oh i see now!

PR #1639, opened 2019-11-28
SPIKE: Support dropping into mixed-size droppable areas

danieldelcore:
## What? 🏗
Adds better support for mixed-size droppable areas. 

## Why? 
When large draggables are dragged over smaller droppable areas, there is some weirdness with the selection of the drop zone. 

**Before:**
image

**After:**

image

### DIFF
--- a/src/state/get-droppable-over.js
+++ b/src/state/get-droppable-over.js
@@ -1,22 +1,66 @@
 // @flow
-import type { Position, Rect } from 'css-box-model';
+import type { Position, Rect, BoxModel } from 'css-box-model';
 import { toDroppableList } from './dimension-structures';
-import isPositionInFrame from './visibility/is-position-in-frame';
-import { find } from '../native-with-fallback';
+
 import type {
   DroppableDimension,
+  DraggableDimension,
   DroppableDimensionMap,
-  DroppableId,
 } from '../types';
+
+const isWithin = (
+  lowerBound: number,
+  upperBound: number,
+): (number => boolean) => (value: number): boolean =>
+  lowerBound <= value && value <= upperBound;
+
+const isPositionIntersecting = (frame: Rect) => {
+  const isWithinVertical = isWithin(frame.top, frame.bottom);
+  const isWithinHorizontal = isWithin(frame.left, frame.right);
+  const frameCenter = frame.center.x;
+
+  return (point: Position, draggableDimensions: BoxModel) => {
+    const { width: draggableWidth } = draggableDimensions.contentBox;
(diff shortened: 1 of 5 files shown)


src/state/get-droppable-over.js
```diff
@@ -1,23 +1,60 @@
...
+  const { width: draggableWidth } = client.borderBox;
+  const activeCenterHorizontal = active.center.x;
+  const isCollidingLeft =
+    activeCenterHorizontal > target.x - draggableWidth / 2;
+  const isCollidingRight =
+    activeCenterHorizontal < target.x + draggableWidth / 2;
+
+  // TODO: shrink active
+
+  // TODO: current draggable rect have any overlap with smaller active
```
danieldelcore: This is where it's falling over currently. We need to figure out how to detect if the active `draggable` has any overlap with a `droppable`. Currently when we drag out of bounds (below the droppable's hit-zone) it will still be selected because it passes the first two edge detection cases. 

I feel compelled to check for an intersection but i'm wondering if there's a better way to do it. Also, we'll need to pass more information through as `active` since it's currently a center point (x, y), we will need `top`, `left`, `right`, `bottom` to calculate the intersection. Shouldn't be too hard to do if you think that's a good idea.

danieldelcore: We might be able to do-away with the edge-detection stuff below if we do that, come to think of it....

src/state/get-droppable-over.js
```diff
@@ -1,23 +1,60 @@
...
+
+  // Has one of the edges passed a center point
+
+  // TODO: check vertical cases
```
danieldelcore: These two checks dont consider vertical cases, where a `draggable` is being moved upward from another `draggable` below it.

src/state/get-droppable-over.js
```diff
@@ -29,13 +66,17 @@ export default ({ target, droppables }: Args): ?DroppableId => {
...
   );
 
-  return maybe ? maybe.descriptor.id : null;
+  const origin = droppables[draggable.descriptor.droppableId];
+
+  const preferences = active.sort(
+    (prev, next) =>
+      distance(origin.client.borderBox.center, next.client.borderBox.center) -
+      distance(origin.client.borderBox.center, prev.client.borderBox.center),
+  );
```
danieldelcore: Preferences are based on distance from the originating `droppable` center point. This might have issues for elements that are dragged from the top of one `droppable` and down to the bottom of another really long `droppable`...

alexreardon:
Absorbed into #1643

PR #1643, opened 2019-11-29
New logic for moving between lists

alexreardon:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-beautiful-dnd",
-  "version": "12.1.1",
+  "version": "12.2.0-alpha.1",
```
danieldelcore: Might want to make sure we remove this

test/unit/state/get-droppable-over/preferencing.spec.js
```diff
@@ -0,0 +1,118 @@
...
+  descriptor: {
+    id: 'my draggable',
+    index: 0,
+    type: droppableLarge.descriptor.type,
+    droppableId: droppableLarge.descriptor.id,
+  },
+  borderBox: droppableLarge.client.borderBox,
+});
+
+it('should prefer lists that are further away in the case that multiple lists are hit', () => {
```
alexreardon: @danieldelcore can you please add a comment explaining which two cases are being hit?

package.json
alexreardon: It's alright - it is going to `dev`

alexreardon: and it is fine to go into `master` with that. Master != production right now

PR #1657, opened 2019-12-04
Fix the distance calculation to support varied lists 

danieldelcore:
In cases where lists are extremely tall (in relation to their siblings) we need to make sure preferencing based on distance between center points still holds up.


alexreardon:
🤘

PR #1675, opened 2019-12-19
Added codesandbox to examples in docs

zrachlin:
Hi! I'm a novice at contributing to open source, so I apologize if I'm not doing this the right way, but I wanted to share this example with the community. I basically combined the "Simple DnD between two lists" example with the "Using with function components" example, and then added the ability to add a droppable area and add/delete an item into a new droppable area. It's not much, but I think it shows some of the awesome modularity and power of react-beautiful-dnd.

Happy to remove this PR and add the suggestion elsewhere if you let me know the best way to proceed. Or just ignore this and I'll continue on my merry way!

Thanks for the amazing library!! It's exactly what I was looking for.


alexreardon:
thanks for contributing this back!

PR #1692, opened 2019-12-26
Update tables.md

elamje:
Added a beginner friendly example that doesn't require using responders. I specifically had a lot of trouble getting the styling (column widths) to apply while an element was being dragged using the current fixed layout strategy. The current guide makes it sound like you must use event responders and set styling based on if the element is dragging or not. This is not necessary (see sandbox link), and adds complexity for many beginners.


alexreardon:
This is great, thank you

alexreardon:
We will get this merged asap. It looks like our build is having some issues

PR #1705, opened 2020-01-13
Export default sensors

### LINKED ISSUE #1698: Please export the default sensors from the npm package
### Description
I am creating a drag-and-drop interaction where the end users will either use a mouse or a touch screen. Keyboard/screenreaders have a different interaction where they get to click the destination lists to move the item from the starting position to the ending position. I would like a way to basically just implement the 2 sensor APIs that I want, while leaving the Keyboard sensor API effectively disabled. But I can't do that the way the NPM package is published. The sensor hooks are not part of the export, and there isn't a way to import them from a standalone file either.

e.g. `<DragDropContext sensors={[useMouseSensor, useTouchSensor]} enableDefaultSensors={false} />`

nerdkid93:
Fixes #1698

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,25 +1,25 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 378196,
-    "minified": 139537,
-    "gzipped": 42095
+    "bundled": 380612,
+    "minified": 139921,
+    "gzipped": 42073
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 324170,
-    "minified": 117548,
-    "gzipped": 35253
+    "bundled": 322226,
+    "minified": 115356,
+    "gzipped": 34189
   },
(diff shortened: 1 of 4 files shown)


alexreardon:
This is great! Can you please update the `drag-drop-context.md` docs to include information about this?

fnpen:
@alexreardon I have big problems with app keyboard handlers together with drag&drop, I need to disable it,
What do we need to fix to release it? I want to help. The only information in `drag-drop-context.md`?

tedgoddard:
It's great to see this PR. I have a custom keyboard sensor (which could maybe even extend the existing keyboard sensor with it exported) so want to disable the existing keyboard sensor but, of course, keep the mouse sensor.

danieldelcore:
Hey @nerdkid93, we just merged in [this pr](https://github.com/atlassian/react-beautiful-dnd/pull/1838) which is based on yours. I'm going to close this PR but I just wanted to say a big thank you for contributing!

PR #1711, opened 2020-01-16
Adding interactive elements tagging functionality

uripre:
Added support for a data-interactive-element property, to mark elements inside Draggables as interactive elements, even if they are not truly HTML-editable.

### DIFF
--- a/docs/api/draggable.md
+++ b/docs/api/draggable.md
@@ -453,6 +453,7 @@ It is possible for your `<Draggable />` to contain interactive elements. By defa
 - `video`
 - `audio`
 - [`contenteditable`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable) (any elements that are `contenteditable` or are within a `contenteditable` container)
+- Any elements that have a `data-interactive-element` property set to true, or any element that is contained in such element.
 
 You can opt out of this behavior by adding the `disableInteractiveElementBlocking` prop to a `<Draggable />`. However, it is questionable as to whether you should be doing so because it will render the interactive element unusable. If you need to _conditionally_ block dragging from interactive elements you can add the `disableInteractiveElementBlocking` prop to opt out of the default blocking and monkey patch the `dragHandleProps (DragHandleProps)` event handlers to disable dragging as required.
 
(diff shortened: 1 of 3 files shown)


docs/api/draggable.md
```diff
@@ -453,6 +453,7 @@ It is possible for your `<Draggable />` to contain interactive elements. By defa
 - `video`
 - `audio`
 - [`contenteditable`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/contenteditable) (any elements that are `contenteditable` or are within a `contenteditable` container)
+- Any elements that have a `data-interactive-element` property set to true, or any element that is contained in such element.
```
alexreardon: can you add one to `data-attributes.js`

Something like:

`${prefix}-is-interactive-element`

Lets have more of a think about the name

Otherwise this looks good!

uripre: Hi Alex, thanks for the feedback!

Could you check if the changes I made are the ones you had in mind?

alexreardon: I'll try to take a look tomorrow!

src/view/use-sensor-marshal/is-event-in-interactive-element.js
```diff
@@ -43,6 +44,12 @@ function isAnInteractiveElement(parent: Element, current: ?Element) {
     return true;
   }
 
+  const dataAttributeName: string = interactiveElement.base;
+  const dataAttribute: ?string = current.getAttribute(dataAttributeName);
+  if (dataAttribute === 'true' || dataAttribute === '') {
```
alexreardon: this feels right, but just want to question it. So we want it for either true or empty string? What if it is just `data-rbd-is-interactive-element` 🤔

uripre: `data-rbd-is-interactive-element` by itself means data-rbd-is-interactive-element={true}, so I think it makes sense.

Empty string is questionable, but I opted for the same behavior as for the existing HTML contenteditable attribute.

alexreardon: Good thinking. I will take a look at this next week

uripre: @alexreardon Hi Alex. Got any time to review it? :)

sguergachi:
Hi @danieldelcore any idea if this will be merged any time soon?

PR #1716, opened 2020-01-22
trying to fix lift announcing

### LINKED ISSUE #1695: VoiceOver in Chrome is not announcing lift instructions
### Expected behavior
VoiceOver announces the default lift instructions: "Draggable item. Ensure your screen reader is not in browse mode and then press space bar to lift.".

### Actual behavior
VoiceOver does not announce the lift instructions

### Steps to reproduce
1. Open https://react-beautiful-dnd.netlify.com/?path=/story/accessibility--single-list in Chrome
2. Turn on VoiceOver
3. Tab to one of the draggable items

### Suggested solution?
Add a default role to the draggable items that is an inter-actable role.

The reason I would fix it this way is due to the following:
image
which comes from https://www.w3.org/TR/using-aria/#label-support

I would probably use the "button" role, but I believe any of the Widget roles will work. https://www.w3.org/TR/wai-aria/#widget_roles

I see this in my implementation on version 16.12.0, but it is also occurring on whatever version is being used by your implementation I referenced in Steps to reproduce.

I see this in my implementation on version 12.2.0, but it is also occurring on whatever version is being used by your implementation I referenced in Steps to reproduce.

Chrome Version 79.0.3945.88 (Official Build) (64-bit)

alexreardon:
Fixes #1695 

TODO:

- [ ] add docs for `role`: why we add it, when to remove it, when to change it
- [ ] write release notes
- [ ] test and research some more


alexreardon:
Closing in favour of #1741

Issue #1629, opened 2019-11-24
core-js 2 as dependency


Bielik20:
### Expected behavior

Polyfills should not be included in library, there should be an instruction what is expected to be included in actual app to support given browser.

### Actual behavior

App expects core-js 2 as a dependency. That breaks apps based on core-js 3.

### Steps to reproduce

If your react app polyfills look somewhat like that: 
```
import 'core-js/stable';

import 'regenerator-runtime/runtime';
```
Then compilation will fail.

### Suggested solution?

Remove polyfills from bundle.

### What version of `React` are you using?

`16.10.2`

### What version of `react-beautiful-dnd` are you running?

`11.0.3`

carlosagsmendes:
I'm experiencing the same issue using webpack with core-js 3

carlosagsmendes:
@Bielik20 would you mind sharing your babel settings? I'm trying to reproduce on a webpack 4 installation from scratch

Bielik20:
@carlosagsmendes I am using [nrwl](https://github.com/nrwl/nx) to manage my monorepo. It is default config from there.

alexreardon:
I'm not sure what the best approach for us should be. Any thoughts @TrySound @theKashey

theKashey:
There are two moment - safety and the ease of use. `core-js` 2/3 is a really big problem and the number affected users would grow, so you have to remove it.
However, it should be a breaking change, or __existing installations would break__.

In short:
- don't include polyfills. As well as ponyfills*
- test and document which ones are required for you. You can use `usage-plugin` in debug more to get a list of polyfills which are _potentially_ required by this library. Keep in mind - that list would be different for `core-js2` and `core-js3`
- after knowing which polyfills are really required you might use ponyfills for some of them (like using `lodash` instead of Array.methods)
- no further talks before the list of polyfills is known (I'll provide it in a hour)

alexreardon:
Ideally, we are not pollyfilling anything. That was my understanding is that we are only using ponyfills

theKashey:
From the code - no, you are not. At least it's not documented - https://github.com/atlassian/react-beautiful-dnd/search?q=runtime-corejs2&unscoped_q=runtime-corejs2 and the only expected polyfills are Number.isInteger and Object.values, which could be ponyfilled... but you can't import thst ponyfill from some specific core-js version(though `runtime-corejs2`).

theKashey:
| File | Required |
| ------------- |:-------------:|
[...]

PR #1725, opened 2020-01-28
Drop core-js in favour of custom fallbacks

TrySound:
Ref https://github.com/atlassian/react-beautiful-dnd/issues/1629#issuecomment-572408426

I forbid using es6+ methods I added eslint-plugin-es6. It does not
forbid isInteger but I send PR https://github.com/nkt/eslint-plugin-es5/pull/37.

Looks like we saved 2kb gzipped. And one less commonjs dependency!


.size-snapshot.json
```diff
@@ -1,25 +1,25 @@
...
-    "gzipped": 41966
+    "bundled": 365555,
+    "minified": 134065,
+    "gzipped": 39834
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 321489,
-    "minified": 115006,
-    "gzipped": 34084
+    "bundled": 307334,
```
alexreardon: nice!

package.json
```diff
@@ -56,7 +56,7 @@
     "prepublishOnly": "yarn build"
   },
   "dependencies": {
-    "@babel/runtime-corejs2": "^7.6.3",
+    "@babel/runtime": "^7.6.3",
```
alexreardon: thanks for sorting this!

src/view/assign-visually-hidden-style.js
```diff
@@ -0,0 +1,16 @@
+// @flow
+// https://allyjs.io/tutorials/hiding-elements.html
+// Element is visually hidden but is readable by screen readers
+const assignVisuallyHidden = (style: Object) => {
```
alexreardon: I prefer my functions not to mutate. Are we avoiding Object.assign?

TrySound: Yes. Object.assign need to be polyfilled. Though in this case much easier to get mutation. I think this is ok for this particular function. object assign polyfill would mutate too.

alexreardon: We use spread quite a bit in the library. Wouldn't that be using Object.assign?

TrySound: Spread uses extends helpers which comes from `@babel/runtime` package

alexreardon: But it won’t use that if you use object.assign. Can we import extends from Babel runtime and use that for the assign? 🤔

alexreardon: My ideal is to not have a function modify the passed in object but rather to return a new object (or to do the Object.assign in place) But I am okay to not get hung up on it

package.json
```diff
@@ -70,6 +70,7 @@
     "@babel/core": "^7.6.4",
     "@babel/plugin-proposal-class-properties": "^7.5.5",
     "@babel/plugin-transform-modules-commonjs": "^7.6.0",
+    "@babel/plugin-transform-object-assign": "^7.8.3",
```
alexreardon: just to confirm - this won't polyfill, it is just ponyfill?

alexreardon: It looks fine: https://babeljs.io/docs/en/next/babel-plugin-transform-object-assign.html#example

TrySound: Yes

TrySound:
You will be able to get rid from eslint suppression in new release of es5 plugin

alexreardon:
Thanks for this @TrySound. I will aim to release this on monday

davidchenfitzgerald:
Hey, when I re added react-beautiful-dnd, I still got the dependency in my lock file for corejs2. Did this merge completely remove it or is there something else I need to do?

davidchenfitzgerald:
Are you guys still planning on a release with this soon?

TrySound:
@davidchenfitzgerald Yes, after fixing a few minor issues

Issue #1679, opened 2019-12-20
[Virtualized] [Android] Scroll instead of drag


neooleg:
### Expected behavior

D&D should start and finished correctly

### Actual behavior

Item partially drags, but mostly scrolls

### Steps to reproduce

1. Open the [example for virtualized](https://react-beautiful-dnd.netlify.com/?path=/story/virtual-react-virtualized--board) on Android 5.0 device in Chrome 79
1. Scroll vertically a bit
1. Try to drag & drop

### What version of `React` are you using?

Official example used

### What version of `react-beautiful-dnd` are you running?

Official example used

### What browser are you using?

Chrome 79 on Android 5.0

### Demo

[Scroll then drag](https://thumbs.gfycat.com/MerrySparseBillygoat-max-1mb.gif)

dsbrlt2002:
Implementation with react-window has the same issue in Google Chrome 79 on Android 5

neooleg:
> [Architectural/Conceptual] Do NOT remove draggable element, but hide it (as it was done in https://github.com/clauderic/react-sortable-hoc).

@alexreardon few architecture questions here, could you please clarify:
1. Is this fit the philosophy of the rbd architecture?
2. Would you accept PR with "if (isiOS) {}"

alexreardon:
I’ll take a look at this soon

neooleg:
@alexreardon did you have a chance to look.

There is fix using this approach:

> Would you accept PR with "if (isiOS) {}"

Do you need a PR for this? Would you accept?

alexreardon:
If we need to do an isIOS, then that is what we need to do! PR would be great

neooleg:
@alexreardon could you please check the PR https://github.com/atlassian/react-beautiful-dnd/pull/1749

lovelle-cardoso:
I'm experiencing this issue too. Any update on this @alexreardon ?

PR #1749, opened 2020-02-12
#1679 fixed

dsbrlt2002:
Issue with drag&drop was found because of different behavior of browser on Android and Ios device
It was fixed by using conditional behaviour with helper function is Ios, which checks user agent


dsbrlt2002:
@alexreardon this is a simple fix for platform specific issue #1679 
could you review it?

Issue #1751, opened 2020-02-16
cleanup settimeout throws error if dom element isn't on page.


sdb1228:
### Expected behavior
When the rbd-announcement-1 div doesn't exist it doesn't try to remove it.

### Actual behavior
rbd tries to remove a dom element that doesn't exist.

```
react-beautiful-dnd.esm.js:5466 Uncaught DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
    at remove (http://localhost:3000/packs/js/application-12c072486af5cffa7be8.js:119874:26)
```

### Steps to reproduce
I haven't gotten an isolated setup (sorry!) but if you use something like turbo links https://github.com/turbolinks/turbolinks with react-beautiful-dnd when you navigate to another page via a link the dom changes before reactdnds settimeout has time to run causing a dom error.

### Suggested solution?

in the root component of your app you can do something like 
```
    window.addEventListener('turbolinks:before-render', () => {
      const highestTimeoutId = setTimeout(';');
      for (let i = 0; i < highestTimeoutId; i++) {
        clearTimeout(i);
      }
    });
```
which will remove the settimeout for reactbeautifuldnd essentially squashing the error.

yes! in `src/view/use-announcer/use-announcer.js:46` add a check before to see if the dom element exists before removing.

### What version of `React` are you using?
16.12.0

### What version of `react-beautiful-dnd` are you running?
12.2.0

### What browser are you using?
chrome Version 79.0.3945.130 (Official Build) (64-bit)

sdb1228:
🎉 Just updated my package.json with the beta version and it silenced this error!  Thanks @alexreardon

PR #1752, opened 2020-02-16
Fix uncaught DOMException error

sdb1228:
This should fix https://github.com/atlassian/react-beautiful-dnd/issues/1751 :)


alexreardon:
When can this happen?

sdb1228:
Hey @alexreardon!  This happens when the dom is cycled out before pretty dnd can unmount (aka this settimeout get executed)

alexreardon:
When did you run into this? I'm guessing in a testing environment?

sdb1228:
Oh! Sorry didn't understand what you meant. I actually ran into this in my development and staging environment on the app I work on. We use Turbo links and sentry had caught this error when someone was navigating away from a react page that had react-beautiful-dnd on it.

alexreardon:
https://github.com/turbolinks/turbolinks ?

alexreardon:
I am thinking we might want to add a test to ensure that we don't break people doing things like this

alexreardon:
For now I'm trying to understand what can cause the issue to occur

sdb1228:
Yup! thats the library.  I can totally add a test.  I can try to find a way to reproduce it.  I was thinking that you could probably experience this if you used something like react-routter  ( https://github.com/ReactTraining/react-router) potentially, but I think the easiest way to trigger this would be to unmount a root level react component at the same time as rendering a full new dom tree so that the timeout happens after the new page is rendered.  I couldn't think of an easy way to reproduce it without spinning up a full rails app with turbo links and everything, but if we really want me to do that I totally can.  In the mean time ill work on writing some tests. :)

alexreardon:
> I think the easiest way to trigger this would be to unmount a root level react component at the same time as rendering a full new dom tree so that the timeout happens after the new page is rendered

Let's add a test for that as I think it is the underlying technique

alexreardon:
Can you please add a test in `unit/integration`?

alexreardon:
(all the tests are moving over to there - all that folder uses react-testing-library)

sdb1228:
Yup! I can totally do that.

src/view/use-announcer/use-announcer.js
```diff
@@ -42,9 +42,11 @@ export default function useAnnouncer(contextId: ContextId): Announce {
         // unmounting after a timeout to let any announcements
         // during a mount be published
         setTimeout(function remove() {
-          // not clearing the ref as it might have been set by a new effect
-          getBodyElement().removeChild(el);
-
+          // checking if element exists before remove to prevent errors
+          if (getBodyElement().contains(el)) {
```
alexreardon: I just checked - this is all good for ie11

https://developer.mozilla.org/en-US/docs/Web/API/Node/contains

alexreardon:
Let me know when you would like me to take a look. Keep to get this in for v13!

sdb1228:
Getting on this this morning! should hopefully have tests up before lunch

sdb1228:
Hey @alexreardon it looks like there is a spec already in `test/unit/view/announcer.spec.js:50` that does exactly what I would need to do just need to modify it a little.  You okay if I just add another spec next to it or would you like me to migrate both tests over to that other directory?

sdb1228:
I pushed up the test and put it in the same file.  It has react testing lib in there already I hope that is okay.  Let me know if you want me to move it.

sdb1228:
And it now has passed all checks.  Let me know if there is anything else you would like me to do!

alexreardon:
I am hoping to release 13.x soon and it would be good to get this in. Any chance you will have an opportunity to look at this soon?

sdb1228:
Thanks for verifying! I can't merge because its a protected branch and I am not authorized to do so.  Never had this happen to me before so let me know what you need me to do on my side @alexreardon

sdb1228:
and I just realized that we are going to be in completely different time zones lol.  Ill stay by my phone to merge this but let me know how I can merge into the branch that is protected.  Ill google around to see if I can figure it out in the mean time.

alexreardon:
> I can't merge

I can do the merging once the PR is ready to go

sdb1228:
>Any chance you will have an opportunity to look at this soon?

Awesome! So maybe I missed it is there some feedback I need to address?

alexreardon:
It looks good, i'll merge it very soon

PR #1754, opened 2020-02-20
Handling early <body> changes

alexreardon:
Follow up #1751 

Adding a test to catch any errors caused by the body changing just before an unmount

/cc @sdb1228


src/view/use-hidden-text-element/use-hidden-text-element.js
```diff
@@ -46,8 +46,11 @@ export default function useHiddenTextElement({
       getBodyElement().appendChild(el);
 
       return function unmount() {
-        // Remove from body
-        getBodyElement().removeChild(el);
+        // checking if element exists as the body might have been changed by thinks like 'turbolinks'
+        const body: HTMLBodyElement = getBodyElement();
```
alexreardon: I found another culprit

test/unit/view/announcer.spec.js
```diff
@@ -59,26 +59,6 @@ it('should remove the element when unmounting after a timeout', () => {
   expect(getElement('5')).not.toBeTruthy();
 });
 
-it('should not remove the element when unmounting after a timeout if the element does not exist', () => {
```
alexreardon: I removed this test in favour of one that catches any issues for all the components

src/view/use-announcer/use-announcer.js
```diff
@@ -41,10 +41,10 @@ export default function useAnnouncer(contextId: ContextId): Announce {
         // unmounting after a timeout to let any announcements
         // during a mount be published
         setTimeout(function remove() {
-          // checking if element exists before remove to prevent errors
-          if (getBodyElement().contains(el)) {
-            // not clearing the ref as it might have been set by a new effect
-            getBodyElement().removeChild(el);
+          // checking if element exists as the body might have been changed by thinks like 'turbolinks'
```
sdb1228: I think you mean things not thinks here :)

src/view/use-hidden-text-element/use-hidden-text-element.js
sdb1228: May want to change to things instead of thinks in the comment :)

test/unit/integration/body-removal-before-unmount.spec.js
```diff
@@ -0,0 +1,37 @@
...
+    getBodyElement().innerHTML = '';
+    unmount();
+    jest.runOnlyPendingTimers();
+  }).not.toThrow();
+
+  jest.useRealTimers();
+});
+
+forEachSensor((control: Control) => {
+  it('should have any errors when body is changed just before unmount: mid drag', () => {
```
sdb1228: 💯 to this test.  Much better sorry I should have probably just migrated it here :)

alexreardon: All good! I wanted to get yours in and see if i could make it catch every component

src/view/use-announcer/use-announcer.js
alexreardon: thanks! changed

PR #1762, opened 2020-02-25
Upgrade react-redux to 7.2.0

timhwang21:
7.1 had some annoying issues like memory leaks, etc. (summarized in
their changelogs). I didn't find a Greenkeeper PR for this for some
reason. Currently we're just using resolutions in our project, but I
figured I might as well file an upstream fix!

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,13 +1,13 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 364497,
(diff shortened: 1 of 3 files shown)


alexreardon:
Looks good to me!

You will just need to run this:

```
yarn bundle-size:update
```

To generate the new bundle size numbers

timhwang21:
Updated -- savings across the board 🎉

PR #1764, opened 2020-02-26
feat: add shouldStartCapture method

chuckWu1989:
1. Support validation before start capture
2. Add validators in app to support more validating functions in the future
3. Add Testing codes
4. Add disable before drag start example

### DIFF
--- a/src/types.js
+++ b/src/types.js
@@ -402,6 +402,8 @@ export type ResponderProvided = {|
   announce: Announce,
 |};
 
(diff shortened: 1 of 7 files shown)


BrenoVelasco:
When is this going to be merged?

chuckWu1989:
I'm ready for review. Require for review

Hivemind9000:
Very keen for this feature to be merged. Using this (brilliant) package in a live multiuser update situation where I need to block the drag while the system renders the view to show updates from other users.

PR #1801, opened 2020-03-31
Add trapped drag example

barabba9174:


alexreardon:
Thanks for this! What is a 'trapped drag'?

barabba9174:
> Thanks for this! What is a 'trapped drag'?

This is an example where dragging outside from the container is avoided

PR #1847, opened 2020-05-24
Fix error in example code

hednowley:


danieldelcore:
Nice one! Thanks a lot!

PR #1890, opened 2020-07-15
Fix getFurthestAway() distance calculation

xian:
Distance should be calculated between the draggable's center and each droppable's center along the major axis.

### DIFF
--- a/src/state/get-droppable-over.js
+++ b/src/state/get-droppable-over.js
@@ -58,8 +58,8 @@ function getFurthestAway({
       const axis: Axis = candidate.axis;
       const target: Position = patch(
         candidate.axis.line,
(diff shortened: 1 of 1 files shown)


xian:
I don't quite understand the assumptions in `getodroppable-over/preferencing.spec' -> `should prefer the second furthest away droppable when multiple lists are hit`.

Without the change, the behavior I'm seeing is that if there 3+ nearby droppables with the same positions and dimensions in the cross axis, the first of them always wins regardless of where the draggable is positioned.

Y0lk:
I came to the exact same solution while trying to figure out why I was having some weird behaviors with multiple droppables. The calculation just seems wrong

xian:
Bump

PR #1894, opened 2020-07-23
Copy edit changes

RavenAlly:
Minor edits, added role and role description notes, added aria-describedby example, added NVDA details.


alexreardon:
Thanks heaps!

PR #1943, opened 2020-09-10
Fix reparenting code example

vkentta:


kobajs (commented):
You may want to add the key property to `<Draggable>` component too

danieldelcore (approved):
Amazing! Thanks!

PR #1981, opened 2020-10-18
Replaced provided.isDragging with snapshot.isDragging in droppable.md file.

agilatakishiyev:
I think this is a mistake, because isDragging property does not exist in DroppableProvided type


agilatakishiyev:
hi there, can you please take a look at this pull request? there is not a big change actually but :disappointed: to see this is open for 15 days already.

danieldelcore (approved):
Great catch! Thanks!

Can confirm based on the `DraggingMapProps` type definition: `src/view/draggable/draggable-types.js`

🙏

PR #1989, opened 2020-10-23
Preventing unnecessary droppable re-renderings

nfantone:
While following the [video tutorial on perf optimizations](https://egghead.io/lessons/react-optimize-performance-in-react-beautiful-dnd-with-shouldcomponentupdate-and-purecomponent), I encountered a few preventable tree re-renderers coming from within `react-beautiful-dnd` itself, thanks to [`wdyr`](https://github.com/welldone-software/why-did-you-render).

1. On _mount_, each `Draggable` on the screen re-sets its state to the same default values in [`getDerivedStateFromProps`](https://github.com/atlassian/react-beautiful-dnd/blob/master/src/view/animate-in-out/animate-in-out.jsx#L35).
2. On _drag_, each `Droppable` placeholder forces an `AnimateInOut` re-render by creating a new `children` function each time.

image

This PR fixes both.


nfantone:
🦗.

nfantone:
🦗 
🦗.

PR #2011, opened 2020-11-10
Rename use-touch-sensor default export to useTouchSensor

ValentinH:
This is not super important but found it misleading that the default exported function name was not matching the actual file name 🙂


danieldelcore:
Love it! Thanks!

PR #2040, opened 2020-12-21
added storysource 5.3.18

2stash:
Adding addon storysource to display source code in storybook examples


2stash:
@danieldelcore This pr requires adding a package. I used npm to install the package. I am still failing the circleci yarn.lock update check because yarn.lock was not updated. I'm not sure how to update the yarn.lock file. I tried to delete yarn.lock and run yarn import, that worked, but I am not sure if that is the right way to update yarn.lock, or if anything was broken.  Please and thank you :)

danieldelcore:
Hey @2stash, with yarn you have to add dependencies via `yarn add [dep-name]`. That will then make sure yarn.lock is updated appropriately. 😀

2stash:
@danieldelcore Happy Holidays! Looks like Australia already hit the new year, is it all sunshine and rainbows in 2021??

Thanks for the info! I didn't think it would be that easy to update both package.json and yarn.lock! The internet said don't have both files so I thought I would just wait for feedback.  All tests have passed, and storysource is able to be displayed. 

I have 2 comments for adding storysource. My assumption is that we want the code to be shown to help developers, especially so they don't post an issue for not finding the source code. (not trying to make this too complicated just trying to be thorough (^O^) ) 

1) Right now the addon panel is not showing by default. When I first saw the react-beautiful-dnd storybook, I did not know what storybook was, so I'm not sure I would have known how to get the addons to show, to see the source code. Should the addon panel show by default? Below is the change needed to get the addons to show by default.

Comment: changing showPanel to true did not do anything.  for addon configurations.
storybook/config.js
```
addParameters({
  options: {
    // currently not using any addons
    showPanel: false, // CHANGING THIS TO TRUE DID NOT SHOW PANEL
    theme,
  },
});
```
Fix: Creating a manager.js or adding the code below to config.js worked, file will show the addon panel by default 
reference; https://github.com/storybookjs/storybook/issues/8383
New File:manager.js or update config.js
```
import { addons } from '@storybook/addons';

addons.setConfig({
  showPanel: true,
  panelPosition: "bottom",
});
```
2) The second issue I see is that the storysource code is not fully sufficient to finding the full code. For example in the first story, "single vertical list, basic", the main react-beautiful-dnd code is in "import QuoteApp from './src/vertical/quote-app'. The full path of this file is "react-beautiful-dnd/stories/src/vertical/quote-app'. So the person looking for the full code will have to still dig around. Some options. 1) do nothing, 2) add a comment to each file about the full path that would be shown in the storysource, 3) add story component that just explains how to use the RBDND storybook.  I did find some examples that tried to address this with code; however the code looked like a lot of work, and would probably break on any updates to storybook, so didn't seem worth the effort.

Thanks again for your help! 
Dan

danieldelcore:
Hey @2stash, 

Thanks again for doing this! The changes look good and I've just pushed a commit that removes the `package-lock` for you!

> 2. The second issue I see is that the storysource code is not fully sufficient to finding the full code.

This is a great point, something I hadn't considered 🤔  I think for now it's ok to merge this in and hopefully it'll provide some value to others. 
But looks like we'll need to rethink how to approach this problem. Rewriting the example code will be a lot of work but I can't think of any other approaches at this time. Any ideas? 🤔

danieldelcore (approved):

2stash:
@danieldelcore Woot! My first PR ever closed! Thanks so much for helping me!  I am still looking for my first software job, so for the next few months I have lots of time to help try and close more issues. This is really fun and great experience! So if you have any other open issues you would support closing please let me know.  I will also go through the open list and try to find some. 

I think the easiest thing would be to update update examples.md. This file links to the examples and is how I found the DND examples, so I would assume that most people would see this page and providing a basic outline of where the files are located would help a lot.  Also some explanation of what stories and storybook are. When I first saw the live example site I immediately thought "omg the site they built to show the DND examples should be a software product on it's own!" 😂

Let me know if you think this is a good path and I can get started, or if you had any other ideas. I haven't thought of anything else that seems worth the effort to implement and maintain. (namely updating each file with a note, or moving stories to their own repo so everything can be at the /src level so files would be in a normal spot)

Issue #1659, opened 2019-12-05
DnD in shadow DOM not working


danielbigler:
### Expected behavior

I´m using rbd in an app that is wrapped by a web component which uses the shadow DOM.

### Actual behavior

Drag and drop is not possible and in the browser console I get the error message, that drag handle could not be found. Is there a way we can get this working?

### Steps to reproduce

Use rbd in the shadow DOM. 

### What version of `React` are you using?

16.12.0

### What version of `react-beautiful-dnd` are you running?
12.1.1

### What browser are you using?

Chrome 78, Firefox 71

### Demo
https://codesandbox.io/s/vertical-list-wn7dd

alexreardon:
We try to find the drag handle using a selector. If the drag handle is within a shadow dom, then my understanding is that we cannot query what is in the shadow DOM from the outside. Does that sound right?

alexreardon:
I could not find a way to do it with a selector. It looks like you need to find the shadow root and then you can query from there. We'll need to think through what is the best way to do this. Ideally, we do not query *every* shadow root.

Some initial thinking:

- ask consumers to put a data attribute on the shadow roots they do want us to query
- allow consumers to pass in a `findDragHandleIdByEvent` override where consumers can do things like query shadow roots

danielbigler:
@alexreardon thank you!

> We try to find the drag handle using a selector. If the drag handle is within a shadow dom, then my understanding is that we cannot query what is in the shadow DOM from the outside. Does that sound right?

Sounds right to me and I think exactly that´s the problem.

> * ask consumers to put a data attribute on the shadow roots they do want us to query

I tried to hack a little solution for this one by attaching the `data-shadow="true"` attribute to the custom element. Afterwards it is possible to find the drag handle by e.g. using
`document.querySelector('[data-shadow="true"]').shadowRoot.querySelectorAll('[data-rbd-drag-handle-context-id="0"]')`.

However, it´s still not possible to move the items in the list. Is it possible, that we also have to fix something with events in the shadow root?

achang97:
Any updates on this? I am also having the same problem.

diegoddox:
I'm facing the same issue as @alexreardon the only thing left is to make the items move.

heyheman11:
I would love to use react-beautful-dnd but can't because of this, it wasn't mentioned in the documentation either

sunshinesanty:
[...]

PR #2264, opened 2021-06-21
Initial proposal to support usage in shadow roots #1659

spielzeugland:
This change is a rough and initial proposal allowing to use this
library at least within the same parent shadow root (but not cross
shadowRoots yet). It does neither cover all functionalities yet nor
is in a quality state to be merged (missing tests etc).

This requires the following aspects
- Make sure not using query selectors on document directly as those
  cannot find elements inside shadow roots. Therefore use specific DOM
  nodes to start searching other nodes
- Events are handled centrally, but event.target will reference to the
  hosting element. Therefore use event.composedPath() to get the real
  nested element for that event.
- Shadow roots isolate stylesheets, means global styles do not bleed
  into the shadow root. Therefore we need to make sure to generate the
  styles into the shadow root. As I couldn't find a way to get the
  mounted DOM node for the DragDropContext component at the point in
  time where the stylesheets are currently getting generated into the
  DOM. Therefore this changes allows consumers to specify the DOM
  location as a prop of the DragDropContext. Same approach is taken by
  JSS (see https://cssinjs.org/jss-api?v=v10.6.0#setup-jss-instance).

### DIFF
--- a/src/view/use-sensor-marshal/use-sensor-marshal.js
+++ b/src/view/use-sensor-marshal/use-sensor-marshal.js
@@ -171,7 +171,7 @@ function tryStart({
     return null;
   }
 
-  const entry: DraggableEntry = registry.draggable.getById(draggableId);
+  const entry: DraggableEntry = registry.draggable.getById(draggableId, sourceEvent.srcElement.shadowRoot || document);
   const el: ?HTMLElement = findDraggable(contextId, entry.descriptor.id);
 
   if (!el) {
--- a/src/view/drag-drop-context/app.jsx
+++ b/src/view/drag-drop-context/app.jsx
@@ -63,6 +63,7 @@ export type Props = {|
 
   // screen reader
   dragHandleUsageInstructions: string,
+  stylesInsertionPoint?: HTMLElement,
 |};
 
 const createResponders = (props: Props): Responders => ({
@@ -107,7 +108,7 @@ export default function App(props: Props) {
     contextId,
     text: dragHandleUsageInstructions,
   });
-  const styleMarshal: StyleMarshal = useStyleMarshal(contextId, nonce);
+  const styleMarshal: StyleMarshal = useStyleMarshal(contextId, nonce, props.stylesInsertionPoint);
 
   const lazyDispatch: (Action) => void = useCallback((action: Action): void => {
     getStore(lazyStoreRef).dispatch(action);
--- a/src/view/get-elements/find-drag-handle.js
+++ b/src/view/get-elements/find-drag-handle.js
@@ -8,10 +8,11 @@ import isHtmlElement from '../is-type-of-element/is-html-element';
 export default function findDragHandle(
   contextId: ContextId,
   draggableId: DraggableId,
+  documentFragment: DocumentFragment
 ): ?HTMLElement {
   // cannot create a selector with the draggable id as it might not be a valid attribute selector
   const selector: string = `[${dragHandleAttr.contextId}="${contextId}"]`;
-  const possible: Element[] = toArray(document.querySelectorAll(selector));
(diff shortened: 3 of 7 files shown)


mrbaguvix (approved):

spielzeugland:
Thanks @mrbaguvix,
I consider your approval as an agreement on the approach itself. I guess there are a lot of other scenarios/places which I haven't found yet as I am not a heavy user of the library myself. Would be great if someone could help me finding those cases e.g. by testing scenarios within shadowRoots. In the meantime there are still failing tests I need to fix.

mrbaguvix:
You can create an issue to this effect maybe the community can better reproduce this.

src/view/use-sensor-marshal/use-sensor-marshal.js
```diff
@@ -171,7 +171,7 @@ function tryStart({
     return null;
   }
 
-  const entry: DraggableEntry = registry.draggable.getById(draggableId);
+  const entry: DraggableEntry = registry.draggable.getById(draggableId, sourceEvent.srcElement.shadowRoot || document);
   const el: ?HTMLElement = findDraggable(contextId, entry.descriptor.id);
```
kehengz, line 175: You added a param "documentFragment" to findDraggable but the caller does not pass in the argument. Change it to findDraggable(contextId, entry.descriptor.id, sourceEvent.srcElement.shadowRoot || document) works for me

kehengz (changes requested):
Thanks a lot for the change! It works now for my use case inside shadowRoot, with a minor change mentioned below :)

src/view/use-sensor-marshal/use-sensor-marshal.js
spielzeugland: Thanks @kehengz, uploaded an additional commit which also fixes this part. The first commit was more of a hacky sketch.
Out of curiosity, did you tested with React 17 or lower?

kehengz: Got it thanks! My repo uses React 17

mrbaguvix (approved):

gorakong:
hey @spielzeugland, we've received your signed CLA. It's just mapped to a different github username (BobPfeiffer?). @danieldelcore (or other maintainers) feel free to merge this while we get the @spielzeugland's entry in our db updated to the correct github account 🙂.

spielzeugland:
Thanks @gorakong and @mrbaguvix for your support - I should have marked this PR as a draft. There is still one TODO left in query-elements.js which I want to address. Probably I should also add some tests before merging?

spielzeugland:
@mrbaguvix I think I am done with the change now. Would be great if you could review it again. As said I worked on allowing shadow roots inside draggables/droppables, added some tests and samples.

Nuurek:
@spielzeugland I've played a bit with your changes - great work, thank you very much for it!

One small fix, in [src/view/use-style-marshal/get-styles.js ](https://github.com/atlassian/react-beautiful-dnd/blob/4b27f9538390668b65632930593575ae883a586d/src/view/use-style-marshal/get-styles.js#L147). There are styles that are supposed to be applied globally using `body` selector. Unfortunately, it won't work in the shadow DOM. It can be fixed by broadening the selector: `selector: 'body, :host'`.

spielzeugland:
@Nuurek for the trying out and suggesting a fix. I uploaded the changes.

spielzeugland:
Hi @danieldelcore, Hi @alexreardon,
would you mind having a look at the changes? From reading your blogs I know bigger changes are difficult to review and accept. Would be happy to get some feedback.

spielzeugland:
Moved to a new PR #2319 with rewritten history in order to satisfy the CLA build.

PR #2270, opened 2021-06-28
Update virtual-lists.md

taqi457:
small typo fix. "So we need you do insert the space for us!" to "So we need you to insert the space for us"

### DIFF
--- a/docs/patterns/virtual-lists.md
+++ b/docs/patterns/virtual-lists.md
@@ -104,7 +104,7 @@ Usually we require consumers to put a `placeholder` (`<Droppable /> | DroppableP
 </Droppable>
 ```
 
-However, a `placeholder` does not make sense in the context of a virtual list as the dimensions of the list is not based on collective size of the visual items, but rather is calculated based on things like `itemCount` and `itemSize`. (eg `height` = `itemSize` \* `itemCount`). For virtual lists, inserting our own node into it would not increase the size of the list. **So we need you do insert the space for us!**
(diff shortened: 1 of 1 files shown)


docs/patterns/virtual-lists.md
```diff
@@ -104,7 +104,7 @@ Usually we require consumers to put a `placeholder` (`<Droppable /> | DroppableP
 </Droppable>
 ```
 
-However, a `placeholder` does not make sense in the context of a virtual list as the dimensions of the list is not based on collective size of the visual items, but rather is calculated based on things like `itemCount` and `itemSize`. (eg `height` = `itemSize` \* `itemCount`). For virtual lists, inserting our own node into it would not increase the size of the list. **So we need you do insert the space for us!**
+However, a `placeholder` does not make sense in the context of a virtual list as the dimensions of the list is not based on collective size of the visual items, but rather is calculated based on things like `itemCount` and `itemSize`. (eg `height` = `itemSize` \* `itemCount`). For virtual lists, inserting our own node into it would not increase the size of the list. **So we need you to insert the space for us!**
```
mrbaguvix, line 107:
```suggestion
However, a `placeholder` does not make sense in the context of a virtual list as the dimensions of the list is not based on collective size of the visual items, but rather is calculated based on things like `itemCount` and `itemSize`. (e.g `height` = `itemSize` \* `itemCount`). For virtual lists, inserting our own node into it would not increase the size of the list. **So we need you to insert the space for us!**
```

mrbaguvix (changes requested):

PR #2285, opened 2021-07-26
Fix removeChild failure similar to #1751

ulion:
I got following error sometimes:
> NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.

And the stack indicate it came from dnd useStyleMarshal.

So made this PR which fix a similar case like #1751, just in the other place of the repo.


princefishthrower:
Getting the same but in `useIsomorphicLayoutEffect`, line 5017 in `react-beautiful-dnd.esm.js`

PR #2297, opened 2021-08-06
Adding axis lock

### LINKED ISSUE #538: support lockaxis feature
Hello guys,

I know a ticket for this has already been opened and closed because for now support for lock axis isn't part of the atlassian use case but it may be a use case of different people (such as my case for example) which prevents me from using this framework making me have to switch to an inferior one.

Would you pretty please consider adding lock axis support?

I'm sure there are other people that ended up not using this framework because of that.

Thank you for your time.

oauo:
Fixes #538

`lockedAxis` and `allowedDeviation` properties on `<Draggable />` restricting dragging to one axis but permitting some movement in the blocked axis.

The main counter point was that this library is meant to feel physical so outright blocking movement in one axis feels like you are interacting with software as not many physical things are "on rails". However sometimes you do just need to lock movement to one axis but given some allowed deviation along the blocked axis (for instance 40px deviation allows -40px to +40px along the blocked axis).

There is a slight problem that I've seen on previous external solutions where keyboard controls are forgotten about, locked axis shouldn't be used as an alternative to `type`.

### DIFF
--- a/src/animation.js
+++ b/src/animation.js
@@ -53,8 +53,22 @@ export const transitions = {
   placeholder: `height ${outOfTheWayTiming}, width ${outOfTheWayTiming}, margin ${outOfTheWayTiming}`,
 };
 
-const moveTo = (offset: Position): ?string =>
-  isEqual(offset, origin) ? null : `translate(${offset.x}px, ${offset.y}px)`;
+const clamp = (value:number, min:number, max:number) =>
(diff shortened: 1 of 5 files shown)


probablyup:
@danieldelcore would love this feature when you get the time to review

MibbitTom:
Bump

PR #2340, opened 2021-11-17
Remove pure: true, for react-redux 8 compatibility

circlingthesun:
When using react-redux 8.0.0-alpha.1 with rbd, you get the following error:

`'The 'pure' option has been removed. 'connect' is now always a "pure/memoized" component'`

Given that `pure: true` was the default before, removing this option ensures that rbd works with react-redux 7 & 8.

### DIFF
--- a/src/view/draggable/connected-draggable.js
+++ b/src/view/draggable/connected-draggable.js
@@ -360,8 +360,6 @@ const ConnectedDraggable = connect(
   {
     // Using our own context for the store to avoid clashing with consumers
     context: StoreContext,
-    // Default value, but being really clear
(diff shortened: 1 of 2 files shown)


davbrito:
Now that react-redux v8 is out this should be accepted

guoguoqiqi:
Still not accepted? 😔

SandeshR98:
Any other work around to this issue?

mikkopori:
Would be nice to get this fixed to avoid the warning in the console.

ericbdev:
It would be great to get this in!

PR #2406, opened 2022-05-19
Adds React 18.1+ to peer dependencies

solsson:
Necessary to use the lib with React 18 through npm 7+ without --legacy-peer-deps


hnrtk:
@solsson is there any way to install the package in react 18.1?

PR #2447, opened 2022-10-04
Supporting use in shadow roots

johnnyfekete:
This commit addresses #1659 and allows use of RBD inside shadow roots e.g. in
Web Components. It does not cover drag&drop between different shadow roots.

The following changes were needed:

- To find the event target nested inside a shadow root we use composedPath
  instead of event.target (since event.target points just to the shadow root
  host).
- To find draggables/droppablesInstead we use query selectors on root node of
  the event source instead of the document. This is encapsulated in a new
  central place (src/view/get-elements/query-elements.js).
- Shadow roots isolate stylesheets, means global styles do not bleed into the
  shadow root. Built-in styles need to be added to the shadow root. This can be
  done by allows consumers to specify the DOM location as a
  stylesInsertionPoint prop on the DragDropContext. Similar approach is taken
  by JSS (see https://cssinjs.org/jss-api?v=v10.6.0#setup-jss-instance).

### DIFF
--- a/.size-snapshot.json
+++ b/.size-snapshot.json
@@ -1,18 +1,18 @@
 {
   "dist/react-beautiful-dnd.js": {
-    "bundled": 364998,
-    "minified": 133574,
-    "gzipped": 39437
+    "bundled": 366336,
+    "minified": 133967,
+    "gzipped": 39652
   },
   "dist/react-beautiful-dnd.min.js": {
-    "bundled": 306810,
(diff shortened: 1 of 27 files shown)


sbahir:
@johnnyfekete I have a similar issue with Shadow DOM unable to find drag context

Does your commit solve this issue? did you find an alternative solution?

Thank you!

PR #2454, opened 2022-11-10
Update draggable.md style spread example

paulmoliva:
Style overrides should occur AFTER the spread of the style props, to match the guide.


danieldelcore (approved):
Great catch! thanks!

PR #2499, opened 2023-04-20
Docs: Update setup-problem-detection-and-error-recovery.md

k-sav:
Fix small typo. 

Assuming a non-rbd /runtime error also _cancels_ any active drag.


danieldelcore:
Thank you!

